PR #4, opened 2019-12-03
Update CHANGELOG.md

ppf2:
Changelog is incorrectly referencing rabbitmq repositories.  Fixed URLs.


kares:
thanks for taking the time to correct these - they're resolved on master, now.

PR #8, opened 2019-12-16
Fix: properly handle empty ssl_endpoint_identification_algorithm

robcowart:
Setting `ssl_endpoint_identification_algorithm` to an empty string disables server hostname verification. In this case the truststore is not needed.

This PR fixes the situation where the `ssl_endpoint_identification_algorithm` was set to an empty string, yet Logstash was failing due to the truststore not being configured.

```
[ERROR][logstash.javapipeline    ][main] Pipeline aborted due to error {:pipeline_id=>"main", :exception=>#<LogStash::ConfigurationError: ssl_truststore_location must be set when SSL is enabled>
```

This change fixes this condition and allows Logstash to connect to Confluent Cloud.


derplarsen:
Thank you! I have tested this and it works like a champ. MUCH appreciated

kares:
thanks - this is looking good, any chance for a test (spec) for the given configuration?

robcowart:
I have no clue how to do that. How would you even test it with out a Kafka environment to which you could attempt to connect?

kares:
Hey Rob, you can do that by stubbing out the internals (in a unit spec) e.g. : 
```diff
diff --git spec/unit/outputs/kafka_spec.rb spec/unit/outputs/kafka_spec.rb
index 5701e79..7266321 100644
--- spec/unit/outputs/kafka_spec.rb
+++ spec/unit/outputs/kafka_spec.rb
@@ -88,6 +88,23 @@ describe "outputs/kafka" do
     end
   end

   context "when a send fails" do
     context "and the default retries behavior is used" do
       # Fail this many times and then finally succeed.
@@ -189,4 +206,25 @@ describe "outputs/kafka" do
       end
     end
   end
+
+  context 'when ssl endpoint identification disabled' do
+
+    let(:config) do
+      simple_kafka_config.merge('ssl_endpoint_identification_algorithm' => '', 'security_protocol' => 'SSL')
+    end
+
+    subject { LogStash::Outputs::Kafka.new(config) }
+
+    it 'does not configure truststore' do
+      expect(org.apache.kafka.clients.producer.KafkaProducer).
+          to receive(:new).with(hash_excluding('ssl.truststore.location' => anything))
+      subject.register
+    end
+
+    it 'sets empty ssl.endpoint.identification.algorithm' do
+      expect(org.apache.kafka.clients.producer.KafkaProducer).
+          to receive(:new).with(hash_including('ssl.endpoint.identification.algorithm' => ''))
+      subject.register
+    end
+  end
 end

```
... `plugin.register` is expected to fail wout your patch. let us know what you think.

unfortunately some parts of the unit specs assume a Kafka setup (as CI does using the scripts in the repository), do not let those confuse you.

robcowart:
Sorry, but I have never looked at the testing methods in Ruby. I am not a Ruby developer. I just hack things together occasionally with the help of Google. Feel free to add this yourself if you feel it is important enough. I have tested in the real-world (used in multiple environment) since I created the PR, so I have confidence in the fix.

robcowart:
@kares Can you please add me to the list of contributors?

derekha-igs:
I'm sorry but this change makes me cannot add truststore without hostname verification, our root CA is not public, with this change will throw SSL handshake fail due to truststore is missing

colinsurprenant:
As @derekha-igs reported, I think this PR introduced a regression which I describe in #52

PR #10, opened 2020-01-10
Fix links

GoodMirek:


GoodMirek:
I see #3 already tries to address this issue, but has not been merged yet, for over two months.

PR #18, opened 2020-03-12
Test: avoid insist usage in spec 

kares:
... for devutils 2.0 compatibility

also scoped `java_import` to reduce potential name clashes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 10.0.1
   - Fix links in changelog pointing to stand-alone plugin changelogs.
+  - Refactor: scope java_import to plugin class
 
 ## 10.0.0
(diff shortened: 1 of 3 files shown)


kares:
merged: https://github.com/logstash-plugins/logstash-integration-kafka/pull/3 ... so that the refactoring fits for the 10.0.1 minor release

robbavey:
@kares There is a [test failure ](https://travis-ci.org/github/logstash-plugins/logstash-integration-kafka/jobs/661510151?utm_medium=notification&utm_source=github_status)on master that looks like it might be related.

kares:
Thanks Rob, those `NoSuchMethodError` are due the recent JRuby 9.2.11 upgrade.
Dealing with those is planned (by doing a mass plugin upgrade): https://github.com/elastic/logstash/issues/11662

PR #23, opened 2020-03-16
Test: replace old (poseidon) kafka client

kares:
[poseidon](https://github.com/bpot/poseidon) gem is defunct, **ruby-kafka** client just reached 1.0.0 so should be a decent alternative

The motivation here is to make sure nothing is interfering as we'll upgrade Kafka clients lib (to 2.4).

### DIFF
--- a/logstash-integration-kafka.gemspec
+++ b/logstash-integration-kafka.gemspec
@@ -49,6 +49,6 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'logstash-devutils'
   s.add_development_dependency 'rspec-wait'
-  s.add_development_dependency 'poseidon'
+  s.add_development_dependency 'ruby-kafka'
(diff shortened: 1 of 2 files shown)


andsel:
The `ruby-kafka` plugin seems to have limited support for Kafka 2.x version, as in https://github.com/zendesk/ruby-kafka#compatibility, could this be a problem?

kares:
Thanks Andrea, for now we only use the Kafka client for tests.
Checked all available clients and they all declare limited 2.x support (think it only means that they do not implement all Kafka features), still ruby-kafka seems the best maintained one. 
The previously used client https://github.com/bpot/poseidon only declares 0.8 compatibility (but still works with kafka 2.3 as far as the integration tests are concerned).

andsel:
To me could seems ok, but there is travis task that repeatedly fails :-(

PR #25, opened 2020-03-23
Refactor: partition_assignment_strategy to accept enumerated values

kares:
this is follow-up from: https://github.com/logstash-plugins/logstash-integration-kafka/pull/16 where a `partitioner` option got introduced which enumerates a list of supported values and maps them to Java class names. 

a few Kafka customization options accept class names instead of enumerated values. 
IMO this makes the configuration a bit ugly and closely tied to Kafka's client API details.
(e.g. `value_serializer => 'org.apache.kafka.common.serialization.StringSerializer'`)

so here's a review of those, turns out that only one setting `partition_assignment_strategy` supports multiple values wout a default so it has been refactored to accept an enumerated list of values, which are documented, while also allowing for class-names due compatibility..

options such as `key_serializer`, `value_serializer` have a default or do not have much variability, so I've left those as they are for now.

also some docs rendering fixes incorporated (for pending 10.1.0 release)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## 10.1.0
-  - updated kafka client (and its dependencies) to version 2.4.1
+  - updated kafka client (and its dependencies) to version 2.4.1 ([#16](https://github.com/logstash-plugins/logstash-integration-kafka/pull/16))
   - added the input `client_rack` parameter to enable support for follower fetching
   - added the output `partitioner` parameter for tuning partitioning strategy
   - Refactor: normalized error logging a bit - make sure exception type is logged
+  - Refactor: partition_assignment_strategy to accept enumerated values ([#25](https://github.com/logstash-plugins/logstash-integration-kafka/pull/25))
 
 ## 10.0.1
   - Fix links in changelog pointing to stand-alone plugin changelogs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
   - added the output `partitioner` parameter for tuning partitioning strategy
   - Refactor: normalized error logging a bit - make sure exception type is logged
   - Fix: properly handle empty ssl_endpoint_identification_algorithm [#8](https://github.com/logstash-plugins/logstash-integration-kafka/pull/8)
-  - Refactor: partition_assignment_strategy to accept enumerated values ([#25](https://github.com/logstash-plugins/logstash-integration-kafka/pull/25))
+  - Refactor : made `partition_assignment_strategy` option easier to configure by accepting simple values from an enumerated set instead of requiring lengthy class paths ([#25](https://github.com/logstash-plugins/logstash-integration-kafka/pull/25))
 
 ## 10.0.1
   - Fix links in changelog pointing to stand-alone plugin changelogs.
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -374,10 +374,11 @@ we haven't seen any partition leadership changes to proactively discover any new
   * Value type is <<string,string>>
   * There is no default value for this setting.
 
-The class name of the partition assignment strategy that the client uses to
-distribute partition ownership amongst consumer instances. Maps to
-the Kafka `partition.assignment.strategy` setting, which defaults to
-`org.apache.kafka.clients.consumer.RangeAssignor`.
+The name of the partition assignment strategy that the client uses to distribute
+partition ownership amongst consumer instances, supported options are `range`,
+`round_robin`, `sticky` and `cooperative_sticky`. These map to Kafka's corresponding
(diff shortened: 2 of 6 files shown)


docs/input-kafka.asciidoc
```diff
@@ -375,10 +374,11 @@ we haven't seen any partition leadership changes to proactively discover any new
   * Value type is <<string,string>>
   * There is no default value for this setting.
 
-The class name of the partition assignment strategy that the client uses to
-distribute partition ownership amongst consumer instances. Maps to
-the Kafka `partition.assignment.strategy` setting, which defaults to
-`org.apache.kafka.clients.consumer.RangeAssignor`.
+The name of the partition assignment strategy that the client uses to distribute
```
andsel: Do think that :
value can be one of [range, round_robin, sticky and cooperative_sticky]

could be better?

lib/logstash/inputs/kafka.rb
```diff
@@ -340,6 +342,24 @@ def create_consumer(client_id)
     end
   end
 
+  def partition_assignment_strategy_class(partition_assignment_strategy = self.partition_assignment_strategy)
```
andsel: Does it make sense to put the opt param when the method is always invoked without parameter?

kares:
Thanks Anread, will look into addressing your concerns.
> Do you think that a unit test on method partition_assignment_strategy_class could help to avoid future regressions?

I believe the functional (integration) spec we have is sufficient, since Kafka will try to instantiate the class set from configuration. Of course could the spec could be improved to test specifics of each strategy but a unit spec is a bit cumbersome since we could only assert on class names.

CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
 ## 10.1.0
-  - updated kafka client (and its dependencies) to version 2.4.1
+  - updated kafka client (and its dependencies) to version 2.4.1 ([#16](https://github.com/logstash-plugins/logstash-integration-kafka/pull/16))
   - added the input `client_rack` parameter to enable support for follower fetching
   - added the output `partitioner` parameter for tuning partitioning strategy
   - Refactor: normalized error logging a bit - make sure exception type is logged
   - Fix: properly handle empty ssl_endpoint_identification_algorithm [#8](https://github.com/logstash-plugins/logstash-integration-kafka/pull/8)
+  - Refactor: partition_assignment_strategy to accept enumerated values ([#25](https://github.com/logstash-plugins/logstash-integration-kafka/pull/25))
```
yaauie: Since we made this change to make configuration easier, we should lead with that.

```suggestion
  - Refactor : made `partition_assignment_strategy` option easier to configure by accepting simple values from an enumerated set instead of requiring lengthy class paths ([#25](https://github.com/logstash-plugins/logstash-integration-kafka/pull/25))
```

spec/integration/inputs/kafka_spec.rb
```diff
@@ -12,159 +11,133 @@
...
+      t.run
+      wait(timeout).for { queue.length }.to eq(event_count) unless timeout.eql?(false)
+      block_given? ? yield(queue, kafka_input) : queue
+    ensure
+      t.kill
+      t.join(30_000)
     end
   end
+
+  def thread_it(kafka_input, queue)
```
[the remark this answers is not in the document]
kares: correct, :bowing_man:

docs/input-kafka.asciidoc
```diff
@@ -375,10 +374,11 @@ we haven't seen any partition leadership changes to proactively discover any new
...
+partition ownership amongst consumer instances, supported options are `range`,
+`round_robin`, `sticky` and `cooperative_sticky`. These map to Kafka's corresponding
+https://kafka.apache.org/24/javadoc/org/apache/kafka/clients/consumer/ConsumerPartitionAssignor.html[`ConsumerPartitionAssignor`]
+implementations.
```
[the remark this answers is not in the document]
kares: > Are there cases in the wild where users provide their own ConsumerPartitionAssignor implementations?

there probably aren't since we do not allow loading custom .jar files. 
in theory, there might be if users add stuff to class-path themselves.

> Do we need to document that the fully-qualified path to a class may still be used?

this also came up during https://github.com/logstash-plugins/logstash-integration-kafka/pull/16 as a `partitioner => ...` option was added with enumerated values (also accepting class-names). we have the custom class name documented but I think we should remove that part: 

> The configuration also allows setting a partitioner class name
e.g. `org.apache.kafka.clients.producer.UniformStickyPartitioner`.

so just like here I am for NOT documenting the possibility to set a class-name (unless we're requested to support loading custom classes).

docs/input-kafka.asciidoc
kares: Thanks, I've changed to a bullet list instead.

lib/logstash/inputs/kafka.rb
kares: thx, have thought about potentially unit testing the method - that's the original reasoning.
going to refactor (there's another very similar one in output) to not receive any arguments.

PR #29, opened 2020-04-13
Retry sending messages only for retriable exceptions

### LINKED ISSUE #27: Do not retry sending messages that failed with a permanent exception
The Kafka output plugin will retry messages that failed to send regardless of the exception type, i.e. it will even retry for permanent exceptions (such as a `SerializationException`). By default there is [no limit on the number of retries](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-kafka.html#plugins-outputs-kafka-retries) which means that these kinds of failures indefinitely hang the entire Logstash pipeline.

Fortunately, the Kafka client library already provides a [`RetriableException` base class](https://kafka.apache.org/0110/javadoc/org/apache/kafka/common/errors/RetriableException.html) for exceptions that are retriable (such as a [`TimeoutException`](https://kafka.apache.org/0110/javadoc/org/apache/kafka/common/errors/RetriableException.html)). To prevent the output pipeline from hanging, the retry logic should be adjusted to only retry for exceptions that inherit `RetriableException` and drop (+ log a warning) when the exception is not retriable. This change could be put behind a configuration property to avoid a breaking change.

Prevously reported in https://github.com/logstash-plugins/logstash-output-kafka/issues/178, https://github.com/logstash-plugins/logstash-output-kafka/issues/190, https://github.com/logstash-plugins/logstash-output-kafka/issues/193 and fixed in https://github.com/logstash-plugins/logstash-output-kafka/pull/194 (unmerged PR).

praseodym:
Fixes #27

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 10.5.0
+  - Changed: retry sending messages only for retriable exceptions [#27](https://github.com/logstash-plugins/logstash-integration-kafka/pull/29)
 
 ## 10.4.1
   - [DOC] Fixed formatting issues and made minor content edits [#43](https://github.com/logstash-plugins/logstash-integration-kafka/pull/43)
--- a/docs/output-kafka.asciidoc
+++ b/docs/output-kafka.asciidoc
@@ -327,6 +327,15 @@ Kafka down, etc).
 
 A value less than zero is a configuration error.
 
+Starting with version 10.5.0, this plugin will only retry exceptions that are a subclass of
+https://kafka.apache.org/25/javadoc/org/apache/kafka/common/errors/RetriableException.html[RetriableException]
+and
+https://kafka.apache.org/25/javadoc/org/apache/kafka/common/errors/InterruptException.html[InterruptException].
+If producing a message throws any other exception, an error is logged and the message is dropped without retrying.
+This prevents the Logstash pipeline from hanging indefinitely.
+
+In versions prior to 10.5.0, any exception is retried indefinitely unless the `retries` option is configured.
+
 [id="plugins-{type}s-{plugin}-retry_backoff_ms"]
 ===== `retry_backoff_ms` 
 
--- a/lib/logstash/outputs/kafka.rb
+++ b/lib/logstash/outputs/kafka.rb
@@ -256,13 +256,14 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
           # send() can throw an exception even before the future is created.
           @producer.send(record)
         rescue org.apache.kafka.common.errors.RetriableException => e
-          logger.info("KafkaProducer#send() failed, will retry", :exception => e)
+          logger.info("producer send failed, will retry sending", :exception => e.class, :message => e.message)
           failures << record
           nil
         rescue org.apache.kafka.common.KafkaException => e
           # This error is not retriable, drop event
           # TODO: add DLQ support
-          logger.warn("KafkaProducer#send() failed, dropping record", :exception => e, :record_value => record.value)
+          logger.warn("producer send failed, dropping record",:exception => e.class, :message => e.message,
+                      :record_value => record.value)
           nil
         end
       end
@@ -275,13 +276,14 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
           rescue java.util.concurrent.ExecutionException => e
             # TODO(sissel): Add metric to count failures, possibly by exception type.
             if e.get_cause.is_a? org.apache.kafka.common.errors.RetriableException
-              logger.info("KafkaProducer#send() future failed, will retry", :exception => e.get_cause)
+              logger.info("producer send failed, will retry sending", :exception => e.cause.class,
+                          :message => e.cause.message)
               failures << batch[i]
             elsif e.get_cause.is_a? org.apache.kafka.common.KafkaException
               # This error is not retriable, drop event
               # TODO: add DLQ support
-              logger.warn("KafkaProducer#send() future failed, dropping record", :exception => e.get_cause,
-                          :record_value => batch[i].value)
+              logger.warn("producer send failed, dropping record", :exception => e.cause.class,
+                          :message => e.cause.message, :record_value => batch[i].value)
             end
           end
         end
--- a/lib/logstash/outputs/kafka.rb
+++ b/lib/logstash/outputs/kafka.rb
@@ -236,7 +236,7 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
     remaining = @retries
 
     while batch.any?
-      if !remaining.nil?
+      unless remaining.nil?
         if remaining < 0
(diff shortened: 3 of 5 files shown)


lib/logstash/outputs/kafka.rb
```diff
@@ -250,27 +250,35 @@ def retrying_send(batch)
         begin
           # send() can throw an exception even before the future is created.
           @producer.send(record)
-        rescue org.apache.kafka.common.errors.TimeoutException => e
+        rescue org.apache.kafka.common.errors.RetriableException => e
+          logger.info("KafkaProducer#send() failed, will retry", :exception => e)
           failures << record
           nil
-        rescue org.apache.kafka.common.errors.InterruptException => e
```
kares, line 256: :red_circle: change in behavior we need to understand: `InterruptException` were previously retried now they aren't

praseodym: Good point! [`InterruptException`](https://kafka.apache.org/24/javadoc/org/apache/kafka/common/errors/InterruptException.html) gets thrown "if the thread is interrupted while blocked" according to the [KafkaProducer#send() Javadoc](https://kafka.apache.org/24/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html#send-org.apache.kafka.clients.producer.ProducerRecord-org.apache.kafka.clients.producer.Callback-). I don't think this should happen unless the producer is stopped or otherwise interrupted.

lib/logstash/outputs/kafka.rb
```diff
@@ -250,27 +250,35 @@ def retrying_send(batch)
...
-          logger.warn("producer send failed", :exception => e.class, :message => e.message)
-          failures << batch[i]
+        # We cannot skip nils using `futures.compact` because then our index `i` will not align with `batch`
+        unless future.nil?
+          begin
+            future.get
+          rescue java.util.concurrent.ExecutionException => e
+            # TODO(sissel): Add metric to count failures, possibly by exception type.
+            if e.get_cause.is_a? org.apache.kafka.common.errors.RetriableException
+              logger.info("KafkaProducer#send() future failed, will retry", :exception => e.get_cause)
```
kares, line 273:
```suggestion
              logger.info("producer send failed, will retry sending", :exception => e.cause.class, :message => e.cause.message)
```

pretty much the whole `future.get` exception handling section should get aligned with whatever we do above

lib/logstash/outputs/kafka.rb
praseodym: The current behaviour was introduced in https://github.com/logstash-plugins/logstash-output-kafka/pull/151, which doesn't motivate why `InterruptException` is in there.

lib/logstash/outputs/kafka.rb
praseodym: `:message => e.cause.message` as suggested or `:message => e.message` like it was?

praseodym: PTAL

lib/logstash/outputs/kafka.rb
kares: Thanks Mark, from that conversation it seemed like just following producer's [`send` throws](https://kafka.apache.org/0100/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html#send(org.apache.kafka.clients.producer.ProducerRecord,%20org.apache.kafka.clients.producer.Callback))
... now Kafka's `InterruptException` doesn't inherit from `RetriableException` but I think we should keep the backwards compatible behavior esp. unless we clearly understand most cases for potential interrupts are not recoverable.

praseodym: I've explicitly added `InterruptException` back in the retry logic. I'm still not sure this is the best way to go (maybe rethrowing it as an `InterruptedException` would make more sense), but at least the behaviour is backwards compatible now.

praseodym:
@kares Do you see any other blockers to getting this merged? Thanks!

GiedriusS:
@kares any updates on this?

adammike:
This is a big problem for us, generating a ton of noise in our logs. Any ETA?

kares:
LGTM in general, 
some concerns raised on the original PR https://github.com/logstash-plugins/logstash-output-kafka/pull/194,
thus I am going to submit this for the team to review, for a second set of :eyes: to look at.

p.s. we are going to need a version bump, probably at least minor with a changelog entry.

robbavey:
@praseodym Firstly, sincere apologies for the amount of time this and the the previous PR in the old repo have been hanging around for.

I'm ok with including this - it seems futile to keep on trying to send messages that will never be accepted by Kafka - but I think it makes sense to document the change in behavior, in a similar way to how we do in the [Elasticsearch output](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#_retry_policy), particularly with regard to messages that would generate `RecordTooLargeException`. 

cc @jsvd, doc(@karenzone)

karenzone:
Thanks for the ping, @robbavey.  

@praseodym, I'm happy to review the docs when you're ready.  Please let me know if you have any questions, or I can be of assistance. Thanks!

praseodym:
@kares @robbavey @karenzone Thanks! I've rebased this PR and updated the docs + changelog in the most recent commit. Please take a look!

docs/output-kafka.asciidoc
```diff
@@ -323,6 +323,13 @@ Kafka down, etc).
 
 A value less than zero is a configuration error.
 
+This plugin will only retry exceptions that are a subclass of
```
kares, line 326: this is a good start, would also mention that previous plugin versions (<= 10.3.0) kept retrying all errors forever.

praseodym: Aren’t the docs versioned, i.e. previous versions of Logstash docs will not include this new paragraph? Or do you think it’s better to be explicit and add the version number?

kares: Yes they are but it's quite uncommon for someone comparing docs side-by-side.
As noted before, the ES plugin still mentions a considerable change in behaviour on handling exceptions:
https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#_retry_policy

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## unreleased
```
kares, line 1:
```suggestion
## 10.4.0
```

GiedriusS:
Have been running this for about a week for so via my own fork, thank you! Works well and extremely needed.

CHANGELOG.md
praseodym: Done!

docs/output-kafka.asciidoc
praseodym: I've updated the docs to also mention the old behaviour.

praseodym:
> we will also need a version bump at:
> 
> https://github.com/logstash-plugins/logstash-integration-kafka/blob/05ffbed5081cdbb7ec84dd4a3dc346c33b8cd770/logstash-integration-kafka.gemspec#L3

Done!

PR #30, opened 2020-04-14
Changed: config defaults to be aligned with Kafka client defaults

kares:
Work towards less user-confusion since the plugin functions as a thin Kafka client wrapper.
Users frequently have to look at the Kafka client docs to find the "real" default client will be using.
Also some of the plugin's documentation was outdated and gave incorrect (pre client 2.0) hints.

The down-side is we might need to review these defaults when client libraries are updated (they might change). On the other hand this has the upside of the plugin behaving ~ same between major client upgrades (despite potential default changes on Kafka client's end).

Most `:validate => :string` values accepting integer values where normalized to `:number`.

*NOTE: some defaults already changed in Kafka client 2.X and are reflected in the code/docs.*


docs/output-kafka.asciidoc
```diff
@@ -268,15 +268,15 @@ Available options for choosing a partitioning strategy are as follows:
 ===== `receive_buffer_bytes` 
 
   * Value type is <<number,number>>
-  * Default value is `32768`
+  * Default value is `65536` (64KB).
```
kares, line 271: changed default to align with Kafka 2.4 client defaults

docs/output-kafka.asciidoc
```diff
@@ -268,15 +268,15 @@ Available options for choosing a partitioning strategy are as follows:
...
 
 The size of the TCP receive buffer to use when reading data
 
 [id="plugins-{type}s-{plugin}-reconnect_backoff_ms"]
 ===== `reconnect_backoff_ms` 
 
   * Value type is <<number,number>>
-  * Default value is `10`
+  * Default value is `50`.
```
kares, line 279: changed default to align with Kafka 2.4 client defaults

andsel:
LGTM

karenzone:
@kares  I have a docs change coming shortly. Perhaps we can coordinate version bump and publishing.

kares:
@karenzone cool - will add a changelog entry + bump 
but going to wait for you to merge more changes before publishing ...

docs/output-kafka.asciidoc
kares: was wrong, this actually is still the same: 32KB at Kafka's

docs/output-kafka.asciidoc
kares: Importance: | **low**

kares:
> By bringing Kafka client defaults into our input library (whether in documentation or as actual defaults to parameters of the input), we implicitly made commitments to those who use our library about the stability of those defaults. We have effectively said "if you don't give us a value for this option, we will give XXX value to the client library". Users generally do not expect behaviour to change out of under them in minor- or patch-level releases of this plugin or in Logstash.

Thanks, so there's really just one change: `reconnect_backoff_ms` from 10ms to 50ms
Given the recent release with Kafka client 2.4 I would argue its fine as we're aligning the plugin with recommended defaults (also this setting is categorized as low priority).

Honestly, I was also a bit against mirroring Kafka defaults. But if I put myself in the shoes of a user it actually makes sense. Esp. since we already had some of the configuration aligned with Kafka's settings, so we either should have not mirrored any or all, the later made more sense.

> It is clear that since we absorbed the commitment for these value defaults, the Kafka client has changed their default behaviour -- likely to improve the baseline experience of a wider variety of use-cases -- but because we now provide default values in our own implementation, our outdated values are preventing the client library's updated defaults from being used.

I hear you, but I do not think it will be that bad unless we do a 2.X -> 3.X major upgrade.
Most of the important (high priority category) settings are stable.

> If we intend to track along-side Kafka's recommended default values, I think we need to document the volatility of our default values so that users can clearly understand the risk of performing upgrades. We should also somehow tool this project to warn us when we consume updated Kafka client libraries to ensure that we stay reasonably up-to-date with the client's recommendations.

Happy to draft a documentation note.
Good note regarding tooling, I've started manually in the hopes that it's only a few options... :hammer: 
Would be useful but would ship this one wout the tool - given how often client libraries get updated (to major versions) and the stability of Kafka's defaults.

docs/input-kafka.asciidoc
```diff
@@ -71,7 +71,9 @@ inserted into your original event, you'll have to use the `mutate` filter to man
 
 This plugin supports these configuration options plus the <<plugins-{type}s-{plugin}-common-options>> described later.
 
-NOTE: Some of these options map to a Kafka option. See the https://kafka.apache.org/documentation for more details.
+NOTE: Some of these options map to a Kafka option. Defaults usually reflect the Kafka default setting.
```
kares, line 74: @karenzone could you please take a look if this block sounds meaningful, 
was trying to state that the new defaults might change as Kafka's underlying defaults change : 
> our defaults will attempt to follow sane defaults and best-practices for these settings from the kafka consumer.

karenzone: @kares. I like what you have. Consider adding "and might change if Kafka's defaults change" at the end of the note to make the user impact obvious. 

Like this: 
"NOTE: Some of these options map to a Kafka option. Defaults usually reflect the Kafka default setting, and might change if Kafka's defaults change."

PR #43, opened 2020-06-24
Doc:Fix formatting and minor content edits and bump to v10.4.1

karenzone:


kares:
we did push 10.4.0 so this needs a merge/rebase against master and there's new stuff that could use an :eyes:

PR #44, opened 2020-06-29
Add isolation_level to override ConsumerConfig.ISOLATION_LEVEL_CONFIG

timtebeek:
For https://discuss.elastic.co/t/logstash-integration-kafka-support-or-document-support-for-setting-consumer-isolation-level-to-read-committed/239116

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 10.4.0
-  - added `isolation_level` to override `ConsumerConfig.ISOLATION_LEVEL_CONFIG` to allow for `read_committed` [#44](https://github.com/logstash-plugins/logstash-integration-kafka/pull/44)
+ - added the input `isolation_level` to allow fine control of whether to return transactional messages [#44](https://github.com/logstash-plugins/logstash-integration-kafka/pull/44)
 
 ## 10.3.0
   - added the input and output `client_dns_lookup` parameter to allow control of how DNS requests are made
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 10.4.0
+  - added `isolation_level` to override `ConsumerConfig.ISOLATION_LEVEL_CONFIG` to allow for `read_committed`
+
 ## 10.3.0
   - added the input and output `client_dns_lookup` parameter to allow control of how DNS requests are made
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 10.4.0
-  - added `isolation_level` to override `ConsumerConfig.ISOLATION_LEVEL_CONFIG` to allow for `read_committed`
+  - added `isolation_level` to override `ConsumerConfig.ISOLATION_LEVEL_CONFIG` to allow for `read_committed` [#44](https://github.com/logstash-plugins/logstash-integration-kafka/pull/44)
 
 ## 10.3.0
   - added the input and output `client_dns_lookup` parameter to allow control of how DNS requests are made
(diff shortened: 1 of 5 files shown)


timtebeek:
>  Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement):
> [a3a8785](https://github.com/logstash-plugins/logstash-integration-kafka/commit/a3a8785696d8fe1ec091c56329f212f6a24d7da7)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

Since signed; no idea how to trigger a refresh of this check.

timtebeek:
@kares would you mind having a look at this pull request?

lib/logstash/inputs/kafka.rb
```diff
@@ -114,6 +114,11 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
   # `session.timeout.ms`, but typically should be set no higher than 1/3 of that value.
   # It can be adjusted even lower to control the expected time for normal rebalances.
   config :heartbeat_interval_ms, :validate => :number, :default => 3000 # Kafka default
+  # Controls how to read messages written transactionally. If set to read_committed, consumer.poll()
+  # will only return transactional messages which have been committed. If set to read_uncommitted'
+  # (the default), consumer.poll() will return all messages, even transactional messages which have
+  # been aborted. Non-transactional messages will be returned unconditionally in either mode.
+  config :isolation_level, :validate => :string, :default => "read_uncommitted" # Kafka default
```
kares, line 121: seems this can be only two values: `read_committed` and `read_uncommitted` 
... thus we should validate no other value is set

docs/input-kafka.asciidoc
```diff
@@ -315,6 +316,17 @@ consumers join or leave the group. The value must be set lower than
 `session.timeout.ms`, but typically should be set no higher than 1/3 of that value.
 It can be adjusted even lower to control the expected time for normal rebalances.
 
+[id="plugins-{type}s-{plugin}-isolation_level"]
+===== `isolation_level` 
+
+  * Value type is <<string,string>>
+  * Default value is `"read_uncommitted"`
+
+Controls how to read messages written transactionally. If set to `read_committed`, `consumer.poll()` will only return
```
kares, line 325: users might have no idea what `consumer.poll()` means
we should re-phrase and just state that:  
> If set to `read_committed`, polling messages will only return ...

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 10.4.0
+  - added `isolation_level` to override `ConsumerConfig.ISOLATION_LEVEL_CONFIG` to allow for `read_committed` [#44](https://github.com/logstash-plugins/logstash-integration-kafka/pull/44)
```
kares, line 2: too much detail on the internals, users might not care about: `ConsumerConfig.ISOLATION_LEVEL_CONFIG`
```suggestion
 - added the input `isolation_level` to allow fine control of whether to return transactional messages [#44](https://github.com/logstash-plugins/logstash-integration-kafka/pull/44)
```

docs/input-kafka.asciidoc
timtebeek: Thanks for the suggestion; applied and will push shortly! Had taken this verbatim from the Kafka docs, but agree this kinder wording would help users here.

lib/logstash/inputs/kafka.rb
timtebeek: Had missed that validation option. Applied just now (hopefully correctly).

kares:
... the gemspec version bump to 10.4.0 is still missing

timtebeek:
> ... the gemspec version bump to 10.4.0 is still missing

Did not know that was a thing, but it's fixed now! :)

timtebeek:
Thanks! Looking forward to putting this to use. Any idea when it'll be available? :)

PR #46, opened 2020-07-20
Doc:Replace plugin_header file with plugin_header-integration file

karenzone:
Integration plugins need a header that is different from the stand-alone plugin header.
The plugin docs should point to the integration repo rather than the
stand-alone input and output repo. We also add a note that the plugin
is part of an integration rather than stand-alone.

The new header included in this work was added with https://github.com/elastic/logstash/pull/11891 and https://github.com/elastic/logstash/pull/12163 for the Logstash Reference, and https://github.com/elastic/logstash-docs/pull/891 for the Versioned Plugin Reference.

PR #47, opened 2020-07-30
Doc: Update kafka client version across kafka integration docs

karenzone:
Corrects kafka version in kafka-input and kafka-output docs.


karenzone:
NOTE TO ALL:  I'll handle changelog and version bump as part of #46

PR #48, opened 2020-07-31
Doc:Add attribute for kafka client versions

karenzone:
Replaces hard-coded kafka client version with an attribute to make updating it easier

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -1,6 +1,8 @@
 :plugin: kafka
 :type: input
 :default_codec: plain
+:kafka_client: 2.4
+:kafka_client_doc: 24
 
 ///////////////////////////////////////////
 START - GENERATED VARIABLES, DO NOT EDIT!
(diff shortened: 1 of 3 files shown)


docs/input-kafka.asciidoc
```diff
@@ -59,7 +66,11 @@ The following metadata from Kafka broker are added under the `[@metadata]` field
 * `[@metadata][kafka][partition]`: Partition info for this message.
 * `[@metadata][kafka][offset]`: Original record offset for this message.
 * `[@metadata][kafka][key]`: Record key, if any.
-* `[@metadata][kafka][timestamp]`: Timestamp in the Record. Depending on your broker configuration, this can be either when the record was created (default) or when it was received by the broker. See more about property log.message.timestamp.type at https://kafka.apache.org/10/documentation.html#brokerconfigs
+* `[@metadata][kafka][timestamp]`: Timestamp in the Record. 
+Depending on your broker configuration, this can be
+either when the record was created (default) or when it was received by the
+broker. See more about property log.message.timestamp.type at
+https://kafka.apache.org/{kafka_client_doc}/documentation.html#brokerconfigs
```
karenzone, line 73: Yikes!  Here's an oldie!

docs/output-kafka.asciidoc
```diff
@@ -328,9 +340,9 @@ Kafka down, etc).
 A value less than zero is a configuration error.
 
 Starting with version 10.5.0, this plugin will only retry exceptions that are a subclass of
-https://kafka.apache.org/25/javadoc/org/apache/kafka/common/errors/RetriableException.html[RetriableException]
+https://kafka.apache.org/{kafka_client_doc}/javadoc/org/apache/kafka/common/errors/RetriableException.html[RetriableException]
 and
-https://kafka.apache.org/25/javadoc/org/apache/kafka/common/errors/InterruptException.html[InterruptException].
+https://kafka.apache.org/{kafka_client_doc}/javadoc/org/apache/kafka/common/errors/InterruptException.html[InterruptException].
```
karenzone, line 345: This link using `/25/` is from a recent PR.  Seems like we should standardize it, too. WDYT?

karenzone:
@kares  I replaced the hard-coded values for kafka client and the kafka version in links with attributes.  The goal is to have one place to set the values in each file. Then all of the other references will be updated automatically.

docs/output-kafka.asciidoc
kares: :+1: most definitely - for correctness these should point to `/24` till we upgrade the client

karenzone:
Thanks for the review, kares! <3

karenzone:
This PR is included in changelog/version bump for #47

PR #51, opened 2020-08-27
Add support for Confluent Schema registry and avro serdes

### LINKED ISSUE #11: Add support for Confluent's Avro Serializers / Deserializers and the Confluent Schema Registry.
**Request:**  This is an enhancement request to add support for Confluent's Avro Serializers / Deserializers **_and_** the Confluent Schema Registry.  

**Related issues:**  https://github.com/logstash-plugins/logstash-input-kafka/issues/238 and https://github.com/logstash-plugins/logstash-input-kafka/pull/239

**Products/Versions:**  Logstash 7.5.1, Kafka 2.1.0, logstash-codec-avro 3.2.3, and logstash-integration-kafka 10.0.0

**Additional information:** The specific use-case has SerDes in place and requires access to the Confluent Schema Registry, even if we add the [SerDes](https://docs.confluent.io/current/streams/developer-guide/datatypes.html#data-types-and-serialization) jar (example `kafka-streams-avro-serde-x.y.z.jar`) to the classpath, that would not solve the issue.  We need to add support to access the Confluent Schema Registry.

According the [(still open) PR](https://github.com/logstash-plugins/logstash-input-kafka/pull/239), which is open to add [`the ability to use the KafkaAvroDeserializer directly`](https://github.com/logstash-plugins/logstash-input-kafka/issues/238#issuecomment-386139050).  I also found [this comment](https://github.com/logstash-plugins/logstash-input-kafka/pull/239#commitcomment-26311716) where the originator of the PR states:
> We have been running this logstash plugin with my changes here without any problems. The consumer isn’t throwing any exceptions and is able to use the confluent kafka avro deserializer just fine. Have you tried building/running it and are getting exceptions?

andsel:
This PR fixes #11 
Spec: #58 

This PR introduces the following changes:
- change the `./kafka_test_setup.sh` and `./kafka_test_teardown.sh` to start and stop a local instance of Schema Registry
- added integration tests that upload an Avro schema to the local registry
- changed input plugin to accept `schema_registry_url` and configure it in avro serdes Kafka Consumer. Covered with integration test to read data and verify the merge of Avro deserialized object into the Logstash event.
- introduces proxy options directly passed to the underling library
- introduces key and secret options to have basic authentication accessing the schema. (this is a not free feature in Confluent's Schema Registry, so can't be tested)


andsel:
**for testers**
to avoid the download of Kafka and Schema Registry for each run, avoid the delete and download with:

```diff 
diff --git a/kafka_test_setup.sh b/kafka_test_setup.sh
index 934a1a4..d505fe6 100755
--- a/kafka_test_setup.sh
+++ b/kafka_test_setup.sh
@@ -10,12 +10,12 @@ fi
 
 export _JAVA_OPTIONS="-Djava.net.preferIPv4Stack=true"
 
-rm -rf build
-mkdir build
+#rm -rf build
+#mkdir build
 
 echo "Downloading Kafka version $KAFKA_VERSION"
 # broken url curl -s -o build/kafka.tgz "https://mirrors.ocf.berkeley.edu/apache/kafka/$KAFKA_VERSION/kafka_2.11-$KAFKA_VERSION.tgz"
-curl -s -o build/kafka.tgz "https://archive.apache.org/dist/kafka/$KAFKA_VERSION/kafka_2.11-$KAFKA_VERSION.tgz"
+#curl -s -o build/kafka.tgz "https://archive.apache.org/dist/kafka/$KAFKA_VERSION/kafka_2.11-$KAFKA_VERSION.tgz"
 mkdir build/kafka && tar xzf build/kafka.tgz -C build/kafka --strip-components 1
 
 echo "Starting ZooKeeper"
@@ -26,7 +26,7 @@ build/kafka/bin/kafka-server-start.sh -daemon build/kafka/config/server.properti
 sleep 10
 
 echo "Downloading Confluent Platform"
-curl -s -o build/confluent_platform.tar.gz http://packages.confluent.io/archive/5.5/confluent-community-5.5.1-2.12.tar.gz
+#curl -s -o build/confluent_platform.tar.gz http://packages.confluent.io/archive/5.5/confluent-community-5.5.1-2.12.tar.gz
 mkdir build/confluent_platform && tar xzf build/confluent_platform.tar.gz -C build/confluent_platform --strip-components 1
 
 echo "Setting up test topics with test data"
```

build.gradle
```diff
@@ -36,15 +36,51 @@ buildscript {
...
-    compile 'org.lz4:lz4-java:1.6.0'
+    compile('io.confluent:kafka-avro-serializer:5.5.1') {
+        exclude group: 'org.apache.kafka', module:'kafka-clients'
+    }
+    compile('io.confluent:kafka-schema-serializer:5.5.1') {
+        exclude group: 'org.apache.kafka', module:'kafka-clients'
+    }
+    compile 'io.confluent:common-config:5.5.1'
+    compile 'org.apache.avro:avro:1.9.2'
+    compile('io.confluent:kafka-schema-registry-client:5.5.1') { // not open source!
```
yaauie, line 54: 🤔 I'm finding Apache-licensed source here: https://github.com/confluentinc/schema-registry/blob/5.5.x/client/src/main/java/io/confluent/kafka/schemaregistry/client/SchemaRegistryClient.java

build.gradle
```diff
@@ -36,15 +36,51 @@ buildscript {
...
+        exclude group: 'org.apache.kafka', module:'kafka-clients'
+    }
+    compile('org.apache.kafka:kafka_2.12:2.5.1') { // contains kafka.utils.VerifiableProperties used by kafka-schema-registry-client
+        exclude group: 'org.apache.zookeeper', module: 'zookeeper'
```
yaauie, line 58: Can we instead use [`include`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.SourceTask.html#org.gradle.api.tasks.SourceTask:include(groovy.lang.Closure) to explicitly include the `kafka.utils.VerifiableProperties` that we need instead of manually excluding a bunch of unrelated things?

lib/logstash/plugin_mixins/common.rb
```diff
@@ -0,0 +1,90 @@
...
+        base.extend(self)
+        base.setup_schema_registry_config
+      end
+
+      def setup_schema_registry_config
+        # Option to set key to access Schema Registry.
+        config :schema_registry_key, :validate => :string
+
+        # Option to set secret to access Schema Registry.
+        config :schema_registry_secret, :validate => :string
```
yaauie, line 16:
```suggestion
        config :schema_registry_secret, :validate => :password
```

We should wrap secrets in the `LogStash::Util::Password` to keep them from escaping through logs etc. Note that this requires manual unwrapping with `LogStash::Util::Password#value` when we need to use it.

lib/logstash/plugin_mixins/common.rb
```diff
@@ -0,0 +1,90 @@
+module LogStash
+  module PluginMixins
+    module SchemaRegistry
```
yaauie, line 3: In case some other plugin wants to implement a "schema registry", we should namespace this `KafkaAvroSchemaRegistry` or something similarly descriptive.

lib/logstash/plugin_mixins/common.rb
```diff
@@ -0,0 +1,90 @@
...
+            conn.proxy = schema_registry_proxy.to_s
+          end
+          if schema_registry_key and !schema_registry_key.empty?
+            conn.basic_auth(schema_registry_key, schema_registry_secret)
+          end
+          conn.adapter Faraday.default_adapter
+        end
+        begin
+          response = client.get('/subjects')
+        rescue Faraday::ConnectionFailed
```
yaauie, line 60: we should handle other failure scenarios too, because they will likely cascade into harder-to-reason-about errors

e.g., an HTTP 404 here may or may not cause the following `JSON.parse` to fail.

lib/logstash/plugin_mixins/common.rb
```diff
@@ -0,0 +1,90 @@
...
+          raise LogStash::ConfigurationError, 'Schema registry service not reachable'
+        end
+        registered_subjects = JSON.parse response.body
+        expected_subjects = @topics.map { |t| "#{t}-value"}
+        if (expected_subjects & registered_subjects).size != expected_subjects.size
+          undefined_topic_subjects = expected_subjects - registered_subjects
+          raise LogStash::ConfigurationError, "Missed to define subjects in schema registry: #{undefined_topic_subjects}"
```
yaauie, line 67: Log message could be more clear.

```suggestion
          raise LogStash::ConfigurationError, "The schema registry does not contain definitions for required topic subjects: #{undefined_topic_subjects}"
```

lib/logstash/plugin_mixins/common.rb
```diff
@@ -0,0 +1,90 @@
...
+      private
+      def check_for_schema_registry_connectivity_and_subjects
+        client = Faraday.new(@schema_registry_url.to_s) do |conn|
+          if schema_registry_proxy && !schema_registry_proxy.empty?
+            conn.proxy = schema_registry_proxy.to_s
+          end
+          if schema_registry_key and !schema_registry_key.empty?
+            conn.basic_auth(schema_registry_key, schema_registry_secret)
+          end
+          conn.adapter Faraday.default_adapter
```
yaauie, line 56: I don't think we need to explicitly tell it to use its own default.

lib/logstash/plugin_mixins/common.rb
```diff
@@ -0,0 +1,90 @@
...
+        end
+        registered_subjects = JSON.parse response.body
+        expected_subjects = @topics.map { |t| "#{t}-value"}
+        if (expected_subjects & registered_subjects).size != expected_subjects.size
+          undefined_topic_subjects = expected_subjects - registered_subjects
+          raise LogStash::ConfigurationError, "Missed to define subjects in schema registry: #{undefined_topic_subjects}"
+        end
+      end
+
+      def split_proxy_into_host_and_port(proxy_uri)
```
yaauie, line 71: combination parse-and-mutate-multiple-ivars method could be more clear.

~~~ ruby
def split_proxy_host_and_port(proxy_uri)
  return nil unless proxy_uri && !proxy_uri.empty?

  port = proxy_uri.port

  host_spec = ""
  host_spec << proxy_uri.scheme || "http"
  host_spec << "://"
  host_spec << "#{proxy_uri.userinfo}@" if proxy_uri.userinfo
  host_spec << proxy_uri.host

  [host_spec, port]
end
~~~

which could then be used like:

~~~
  @schema_registry_proxy_host, @schema_registry_proxy_port = split_proxy_host_and_port(schema_registry_proxy)
~~~

lib/logstash/plugin_mixins/common.rb
```diff
@@ -0,0 +1,90 @@
...
+          @schema_registry_proxy_host += "://"
+          @schema_registry_proxy_host += proxy_uri.userinfo + "@" if proxy_uri.userinfo
+          @schema_registry_proxy_host += proxy_uri.host
+        end
+      end
+
+      def check_for_key_and_secret
+        if schema_registry_key and !schema_registry_key.empty?
+          if !schema_registry_secret or schema_registry_secret.empty?
+            raise LogStash::ConfigurationError, "Missed to define schema_registry_secret after defining schema_registry_key"
```
yaauie, line 84:
```suggestion
            raise LogStash::ConfigurationError, "Setting `schema_registry_secret` is required when` schema_registry_key` is provided."
```

yaauie (commented):
Overall, this looks like it will work well, but could use some attention to failure scenarios. I've recently been bitten by code that assumes a successful http connection is a valid connection to a specific resource, which resulted in surprising and hard-to-detect errors when the http connection was established to a different kind of resource.

build.gradle
andsel: Right! was a pinpoint I forgot to remove

lib/logstash/plugin_mixins/common.rb
andsel: I'll try, I added because found [this article](https://blog.thecodewhisperer.com/permalink/losing-time-to-faraday) and a related Faraday issues https://github.com/lostisland/faraday_middleware/issues/32 and https://github.com/lostisland/faraday/issues/121

build.gradle
andsel: That method is only for manipulation of file trees not transitive dependency management. 
To avoid transitive dependency resolution we could use `transitive = false` but also with that `kafka_2.12` transitively include group: 'org.scala-lang.modules', module:' scala-java8-compat_2.12'

lib/logstash/plugin_mixins/common.rb
andsel: That problem was fixed in Faraday 1.0 and we use 1.0.1

lib/logstash/plugin_mixins/common.rb
andsel: widened the types of error that the clause can catch

andsel:
Hi @yaauie thanks for precise review! I've update the PR with your suggestions, LMKWDYT.

lib/logstash/plugin_mixins/common.rb
```diff
@@ -0,0 +1,93 @@
+module LogStash
+  module PluginMixins
+    module KafkaAvroSchemaRegistry
+      DEFAULT_SERIALIZER_CLASS = "org.apache.kafka.common.serialization.StringDeserializer"
```
yaauie, line 4: nit: I would move this constant to `LogStash::Inputs::Kafka`, since it has more to do with the output than the schema registry

yaauie (commented):
Functionality-wise, this is looking great 🎉 

I've left a couple nitpicks, but other than that this LGTM.

docs/input-kafka.asciidoc
```diff
@@ -124,6 +124,10 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-sasl_jaas_config>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-sasl_kerberos_service_name>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-sasl_mechanism>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_key>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_proxy>> |<<string,uri>>|No
```
karenzone, line 128:
```suggestion
| <<plugins-{type}s-{plugin}-schema_registry_proxy>> |<<uri,uri>>|No
```

docs/input-kafka.asciidoc
```diff
@@ -124,6 +124,10 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
...
+| <<plugins-{type}s-{plugin}-schema_registry_secret>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_url>> |<<string,uri>>|No
```
karenzone, line 130:
```suggestion
| <<plugins-{type}s-{plugin}-schema_registry_url>> |<<uri,uri>>|No
```

docs/input-kafka.asciidoc
```diff
@@ -528,6 +532,39 @@ http://kafka.apache.org/documentation.html#security_sasl[SASL mechanism] used fo
...
+
+[id="plugins-{type}s-{plugin}-schema_registry_url"]
+===== `schema_registry_url`
+
+* Value type is <<uri,uri>>
+
+This is the URI that points to an instance of https://docs.confluent.io/current/schema-registry/index.html[Schema Registry] service,
+used to manage Avro schemas. When the Schema Registry is configured it's avoided to specify also the <<plugins-{type}s-{plugin}-value_deserializer_class>>.
+On the Schema Registry service must be previously uploaded the Avro schemas that are used to deserialize the data from the specified topics.
+Those schemas must follow a naming convention with the pattern `<topic name>-value`.
```
karenzone, line 567: Please consider this rewording: 

The URI that points to an instance of the
https://docs.confluent.io/current/schema-registry/index.html[Schema Registry]
service, used to manage Avro schemas. 
Be sure that the Avro schemas for deserializing the data from the specified
topics have been uploaded to the Schema Registry service. 
The schemas must follow a naming convention with the pattern `<topic
name>-value`.

Use either the Schema Registry config option or the
<<plugins-{type}s-{plugin}-value_deserializer_class>> config option, but not
both.

karenzone: Please make sure that there is a return between the last line of this description and the next options anchor.

docs/input-kafka.asciidoc
```diff
@@ -641,7 +678,8 @@ The topics configuration will be ignored when using this configuration.
   * Value type is <<string,string>>
   * Default value is `"org.apache.kafka.common.serialization.StringDeserializer"`
 
-Java Class used to deserialize the record's value
+Java Class used to deserialize the record's value. This settings conflicts with <<plugins-{type}s-{plugin}-schema_registry_url>>.
+A custom value deserializer could be used only if not using a Schema Registry.
```
karenzone, line 682:
```suggestion
Java Class used to deserialize the record's value. 

A custom value deserializer can be used only if you are not using a Schema Registry.
Use either the value_deserializer_class config option or the
<<plugins-{type}s-{plugin}-schema_registry_url>> config option, but not both.
```

karenzone (commented):
Thank you for adding these helpful descriptions. I left some comments inline for your consideration.

docs/input-kafka.asciidoc
andsel: Thanks @karenzone for the suggestion, integrated!

fbaligand:
Great news!
Thanks for this new feature, that I was waiting for!

lib/logstash/inputs/kafka.rb
```diff
@@ -278,6 +285,13 @@ def thread_runner(logstash_queue, consumer)
           for record in records do
             codec_instance.decode(record.value.to_s) do |event|
               decorate(event)
+              if schema_registry_url
+                json = LogStash::Json.load(record.value.to_s)
```
bquinart, line 289: Do I understand that the value of the Kafka message that was in Avro has been converted in a json string and is again being parsed from json add the elements in the 'Logstash event object'?
Is it not possible to bypass the json and go from Kafka to 'Logstash event object' directly?
I am not sure how expensive the json parsing is.

bquinart (commented):
Thanks for this. This new feature will make things easier for us.
Could you clarify the approach taken for the Avro decoding?

lib/logstash/inputs/kafka.rb
andsel: The record is an instance of [ConsumerRecord](https://kafka.apache.org/26/javadoc/org/apache/kafka/clients/consumer/ConsumerRecord.html#ConsumerRecord-java.lang.String-int-long-K-V-) and could be potentially deeply nested. Here the desire is to copy all the keys and values from the record structure into the Logstash event. This is a simple way to do it, but if you have any evidence of bottleneck please raise an issue so that we can try another implementaion

PR #59, opened 2020-10-07
Docs: explain group_id in case of multiple inputs

kares:
since `group_id` isn't required to be set we should note that this won't do for multiple inputs in a single pipeline

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -314,7 +314,11 @@ before answering the request.
 
 The identifier of the group this consumer belongs to. Consumer group is a single logical subscriber
 that happens to be made up of multiple processors. Messages in a topic will be distributed to all
-Logstash instances with the same `group_id`
+Logstash instances with the same `group_id`.
+
+NOTE: In cases when multiple inputs are being used in a single pipeline, reading from different topics,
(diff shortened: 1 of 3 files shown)


docs/input-kafka.asciidoc
```diff
@@ -301,7 +301,11 @@ before answering the request.
 
 The identifier of the group this consumer belongs to. Consumer group is a single logical subscriber
 that happens to be made up of multiple processors. Messages in a topic will be distributed to all
-Logstash instances with the same `group_id`
+Logstash instances with the same `group_id`.
+
+NOTE: In cases when multiple inputs are being used in a single pipeline, reading from different topics,
+it's essential to set a different `group_id => ...` for each input. Setting a unique `client_id => ...`
+is also recommended.
```
karenzone, line 308:
```suggestion
NOTE: When you are using multiple inputs in a single pipeline and reading from different topics,
be sure to set a different `group_id => ...` for each input. Setting a unique `client_id => ...`
is also recommended.
```

karenzone (approved):
I offered a wording suggestion inline for your consideration. 
Otherwise, builds cleanly and LGTM! Thanks for adding this important clarification.

Issue #52, opened 2020-08-27
ssl_endpoint_identification_algorithm handling regression introduced in 10.1.0 in output


colinsurprenant:
It looks like #8 introduced a regression, by not setting the `truststore` options anymore when  `ssl_endpoint_identification_algorithm` s empty.

when a cert does not match the hostname, prior to #8 it was possible to set:

```
     security_protocol => "SSL"
     ssl_endpoint_identification_algorithm => ""

     ssl_truststore_location => "..."
     ssl_truststore_password => "..."
     ssl_keystore_location => "..."
     ssl_keystore_password => "..."
     ssl_key_password => "..."
```

but after #8 (in 10.1.0) this fails with

```
[ERROR][org.apache.kafka.clients.NetworkClient][elastiflow] [Producer clientId=producer-1] Connection to node -1 (...) failed authentication due to: SSL handshake failed
```

kares:
:boom: https://github.com/logstash-plugins/logstash-integration-kafka/pull/8 seemed to make sense but I guess if there's users complaining the only thing left is to revert @robcowart's changes

addodelgrossi:
I'm experiencing the same problem, which doesn't happen in the input plugin.

atoom:
We are having the same issue when updating Logstash from version 7.6.1 to 7.9.2. We are using internal self-signed certificates where the CN does not necessarily match the broker hostname so we have set ssl_endpoint_identification_algorithm to "".

As a workaround we have been forced to downgrade the logstash-integration-kafka plugin to version 10.0.0.

Would love to see this fixed soon so that we are not force to run on outdated version of this plugin.

secfalcon:
Experiencing this issue as well - although truststore is configured, it is no longer used and results in SSL handshake errors. Only experienced on Kafka output plugin; Kafka input plugin is not impacted as breaking change mentioned was only made to Kafka output plugin.

Our workaround was to downgrade logstash-integration-kafka to 10.0.1 as well as we cannot enable enable endpoint identification validation at this time to work around the issue.

colinsurprenant:
Will be fixed by #60

PR #60, opened 2020-10-20
Fix: set (optional) truststore when endpoint id check disabled

kares:
Disabling server host-name verification (`ssl_endpoint_identification_algorithm => ""`) should still allow the plugin (output) to set `ssl_truststore_location => "..."` as described in #52. This is no longer possible since #8 (since **10.1.0**).

Resolves https://github.com/logstash-plugins/logstash-integration-kafka/issues/52
Semi-Reverting: https://github.com/logstash-plugins/logstash-integration-kafka/pull/8

### DIFF
--- a/lib/logstash/plugin_mixins/kafka_support.rb
+++ b/lib/logstash/plugin_mixins/kafka_support.rb
@@ -1,0 +1,29 @@
+module LogStash module PluginMixins module KafkaSupport
+
+  def set_trustore_keystore_config(props)
+    props.put("ssl.truststore.type", ssl_truststore_type) unless ssl_truststore_type.nil?
+    props.put("ssl.truststore.location", ssl_truststore_location) unless ssl_truststore_location.nil?
+    props.put("ssl.truststore.password", ssl_truststore_password.value) unless ssl_truststore_password.nil?
+
+    # Client auth stuff
+    props.put("ssl.keystore.type", ssl_keystore_type) unless ssl_keystore_type.nil?
+    props.put("ssl.key.password", ssl_key_password.value) unless ssl_key_password.nil?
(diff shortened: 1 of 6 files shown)


lib/logstash/plugin_mixins/kafka_support.rb
```diff
@@ -0,0 +1,29 @@
+module LogStash module PluginMixins module KafkaSupport
+
+  def set_trustore_keystore_config(props)
+    props.put("ssl.truststore.type", ssl_truststore_type) unless ssl_truststore_type.nil?
```
kares, line 4: prior to #8 output had a [nil check](https://github.com/logstash-plugins/logstash-integration-kafka/commit/e6f7d3a8be6a606fb498a20ad1d21d48546689ac#diff-3ce6e7cf50af34dab48324fd49c5fb514373531921e89015f1974891f5f5ef87R244-R246) 
```ruby
  if ssl_truststore_location.nil?
    raise LogStash::ConfigurationError, "ssl_truststore_location must be set when SSL is enabled"
  end
```
since Kerberos authentication was introduced at: https://github.com/logstash-plugins/logstash-integration-kafka/commit/e6f7d3a8be6a606fb498a20ad1d21d48546689ac#diff-3ce6e7cf50af34dab48324fd49c5fb514373531921e89015f1974891f5f5ef87R244-R246

input did not have such a check and truststore is optional according to the docs regardless of consumer/producer.
(despite the 'revert' this should still address the problematic case presented in https://github.com/logstash-plugins/logstash-integration-kafka/pull/8)

https://github.com/logstash-plugins/logstash-integration-kafka/pull/60/files#diff-886ed388751b8abe78d71e1bbb13450fd499ef48af6d95f06795f6aa8d825398R53

colinsurprenant (approved):
🚀  Great! Thanks for following up on this, nice DRYing too. 
LGTM.

Issue #38, opened 2020-06-18
Input: make record headers available to events


yaauie:
This is a migration of a feature-request from the stand-alone Input Plugin: https://github.com/logstash-plugins/logstash-input-kafka/issues/336

 - Users want to be able to attach kafka record headers to their events.
 - The record header values are implemented as [arbitrary binary data](https://kafka.apache.org/24/javadoc/org/apache/kafka/common/header/Header.html) and are available via [`org.apache.kafka.clients.consumer.ConsumerRecord#headers`](https://kafka.apache.org/24/javadoc/org/apache/kafka/clients/consumer/ConsumerRecord.html#headers--)

Proposal:
 - migrate `decorate_events` option from boolean to `none`/`basic`/`extended` like we did in the RabbitMQ input to avoid incurring performance penalties when the headers are not explicitly requested https://github.com/logstash-plugins/logstash-integration-rabbitmq/pull/14
 - support only string values that are UTF-8 compatible (including 7-bit ASCII), ignoring others

**Fixed by** : https://github.com/logstash-plugins/logstash-integration-kafka/pull/62

RashmiRam:
@yaauie We are in need of this feature and would like to pick this up. Shall I go ahead and pick this up?

andsel:
@yaauie actually the `decorate_events` options is of `boolean` type, if we move it to the tri-state it changes to `string` type, so we introduce a breaking change I think. How could we manege this without annoy existing user's pipeline configuration? 
My proposal is to keep the existing `decorate_events` as boolean, introduce a new `string` option named `decorate_mode` with the 3 values above. The `decorate_events` is translated to the new `decorate_mode`, then we could warn a deprecation log is the configuration contains the old option.
In case the both new and old options are used, we give precedence to the new warning the clashing.
@yaauie WDYT?

yaauie:
>  if we move it to the tri-state it changes to string type, so we introduce a breaking change

I was a bit lazy in just linking to the other ticket and not fully explaining, but I propose we do it in the same non-breaking way, supporting `true` and `false` inputs as deprecated aliases for `basic` and `none`:

> I changed the formerly boolean `metadata_enabled` to instead accept a value from a list:
> 
> * `none`: no metadata (default)
> * `basic`: header and property metadata
> * `extended`: header, property, and payload metadata
> * `false`: deprecated alias for `none`
> * `true`: deprecated alias for `basic`
> 
[...]

PR #62, opened 2020-11-19
add headers key/value read from the message to the generated logstash event

yanayita:
Solves #38


Zico56:
Hello there,

Is there a release scheduled for this development ?

I actually need to parse headers from kafka message in logstash.
I tested this development with a custom compile of the plugin, it seems to work pretty well.

However, if testing a custom plugin is possible on our dev environement,
for production purpose, I'll prefer to have this in an official release.

Thanks.

yanayita:
Hi Zico, I didn't have time to progress on this. The problem is that CI checks have failed, because in order to implement testing for the headers, we cannot use kafka-console-producer, and instead I relied on using kafcat which allows generating messages with headers for the tests, but cannot use kafcat (using docker) with the current CI setup.

Maybe someone else could advise a workaround? 
Otherwise for now, I could simply remove the tests for this implementation so that CI can complete

andsel:
Hi @yanayita thank you for contributing with this PR. I think we are almost there, but it only needs to address a couple of things:

- instead to introduce another flag (name it `decorate_headers`) I would go with Ry suggestion to have a tri-state `decorate_events`
-  add some testing (unit or integration) to prove the fix
-  the decode phase should be in UTF-8, `header.value.to_s` convert in UTF-8 but it could raise error (this has to be verified) if the value is not a valid UTF-8 string, while in issue #38 was asked to skip in case of parsing error of the single header

andsel:
I close and reopen to see if the webhook to Travis.com is updated

PR #71, opened 2021-02-08
Avoid usage of SHUTDOWN event in Logstash 8.0.0 or greater

andsel:
With PR https://github.com/elastic/logstash/pull/12517 the `SHUTDOWN` event has been removed, this PR avoid to use it if the plugin is running on Logstash 8.0.0 or greater

### DIFF
--- a/lib/logstash/outputs/kafka.rb
+++ b/lib/logstash/outputs/kafka.rb
@@ -224,7 +224,6 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
     end
 
     events.each do |event|
-      break if event == LogStash::SHUTDOWN
       @codec.encode(event)
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 10.7.1
+  - Fix: dropped usage of SHUTDOWN event deprecated since Logstash 5.0 [#71](https://github.com/logstash-plugins/logstash-integration-kafka/issue/71)
+  
 ## 10.7.0
   - Switched use from Faraday to Manticore as HTTP client library to access Schema Registry service 
     to fix issue [#63](https://github.com/logstash-plugins/logstash-integration-kafka/issue/63) 
--- a/logstash-integration-kafka.gemspec
+++ b/logstash-integration-kafka.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = 'logstash-integration-kafka'
-  s.version         = '10.7.0'
+  s.version         = '10.7.1'
   s.licenses        = ['Apache-2.0']
   s.summary         = "Integration with Kafka - input and output plugins"
   s.description     = "This gem is a Logstash plugin required to be installed on top of the Logstash core pipeline "+


kaisecheng:
This is a fix for ruby engine removal. 
We removed `LogStash::SHUTDOWN` in Logstash 8 and confirm the class is not referenced by any Logstash core.
I am not sure will user go into a situation that running Logstash 8 and using an old kafka plugin pointing `if event == LogStash::SHUTDOWN`, if so, the missing class will block the plugin to run.
Can @robbavey have a look?

lib/logstash/outputs/kafka.rb
```diff
@@ -224,7 +224,7 @@ def multi_receive(events)
     end
 
     events.each do |event|
-      break if event == LogStash::SHUTDOWN
+      break if LOGSTASH_VERSION < '8.0' and event == LogStash::SHUTDOWN
```
robbavey, line 227: I'm not sure we need this line of code at all - AFAICT, the `LogStash::SHUTDOWN` event being passed to down to plugins is long removed from Logstash code (https://github.com/elastic/logstash/issues/8847#issuecomment-352573320 refers to it being removed in Logstash `5.0`). WDYT @jsvd?

However... we should also check how prevalent this check is within our plugins - I found one [other](https://github.com/logstash-plugins/logstash-output-nagios_nsca/blob/c81f81ea8decfdee73f5c7cd99454b51d2ad4ccd/lib/logstash/outputs/nagios_nsca.rb#L69) case where this check was made.

robbavey (changes requested):
I don't think we need this check here - the `SHUTDOWN` event being pushed down to plugins has long gone. However, we should double to check to figure out where else we are doing this check, and see what else may be broken

lib/logstash/outputs/kafka.rb
andsel: I checked in all repositories inside `logstash-plugins` organization and for tier 1 and 2 it is used only in the `integration-kafka` and `output-kafka` (which is now deprecated) the rest of references are in other plugins, here is the summary:
```
Tier 1
logstash-integration-kafka/lib/logstash/outputs/kafka.rb:227
logstash-output-kafka/lib/logstash/outputs/kafka.rb:218:       


logstash-filter-collate/lib/logstash/filters/collate.rb
logstash-output-cloudwatch/lib/logstash/outputs/cloudwatch.rb:177:    
logstash-output-jira/lib/logstash/outputs/jira.rb:125:     
logstash-output-lumberjack/lib/logstash/outputs/lumberjack.rb:58:     
logstash-output-nagios_nsca/lib/logstash/outputs/nagios_nsca.rb:69:     
logstash-output-newrelic/lib/logstash/outputs/newrelic.rb:130:    
logstash-output-redmine/lib/logstash/outputs/redmine.rb:115:     
logstash-output-zookeeper/lib/logstash/outputs/zookeeper.rb:49:    
```

given the comment @robbavey  reported, I think we could completely remove the check.

andsel: If we remove it we are backward and forward compatible

jsvd: +++ on removing this check, of the tier 2 I'd only prioritize lumberjack and cloudwatch outputs, as they're included in the logstash distribution

andsel: Removed, PRs for the other two plugins:
- lumberjack output https://github.com/logstash-plugins/logstash-output-lumberjack/pull/31
- cloudwatch output https://github.com/logstash-plugins/logstash-output-cloudwatch/pull/18

jsvd (approved):
LGTM

PR #78, opened 2021-03-15
Feature/read headers

andsel:
This a clone of PR #62 to force TravisCI.com to build it


spec/integration/inputs/kafka_spec.rb
```diff
@@ -91,6 +99,55 @@
...
+        event = queue.shift
+        expect(event.get("[@metadata][kafka][topic]")).to eq("logstash_integration_topic_plain_with_headers")
+        expect(event.get("[@metadata][kafka][consumer_group]")).to eq(group_id_3)
+        expect(event.get("[@metadata][kafka][timestamp]")).to be >= start
+        expect(event.get("[@metadata][kafka][headers][name]")).to eq("John")
+      end
+    end
+
+    it "should skip headers not encoded in UTF-8" do
+      invalid = "😵".to_java_bytes
```
yaauie, line 121: Can you explain why this string is invalid? To me, it appears like you are encoding a UTF-8 emoji so the description of the spec leads me to believe it should work.

The below will report itself as UTF-8, but it contains an invalid byte sequence.

```suggestion
      invalid = "日本".encode(:Shift_JIS).force_encoding('UTF-8')
```

lib/logstash/inputs/kafka.rb
```diff
@@ -292,14 +319,24 @@ def thread_runner(logstash_queue, consumer)
...
                 event.set("[@metadata][kafka][partition]", record.partition)
                 event.set("[@metadata][kafka][offset]", record.offset)
                 event.set("[@metadata][kafka][key]", record.key)
                 event.set("[@metadata][kafka][timestamp]", record.timestamp)
               end
+              if @metadata_mode.include?(:headers)
+                for header in record.headers do
+                  s = String.from_java_bytes(header.value)
+                  begin
+                    s.encode!("UTF-8")
```
yaauie, line 334: Minor optimization: use the `Encoding` instead of a String reference to save a per-header lookup

```suggestion
                    s.encode!(Encoding::UTF_8)
```

yaauie, line 334: I'm not 100% familiar with the implementation of JRuby's `String::from_java_bytes`, but `String#encode!` is a pass-through errorless operation when the input is tagged as being in the destination encoding -- whether or not the underlying byte sequence is valid in that encoding.

> ~~~
> irb(main):001:0> shift_jis = "日本".encode('Shift_JIS')
> => "\x{93FA}\x{967B}"
> irb(main):002:0> shift_jis.valid_encoding?
> => true
> irb(main):003:0> shift_jis.force_encoding('UTF-8') # mutates encoding tat but not bytes
> => "\x93\xFA\x96{"
> irb(main):004:0> shift_jis.valid_encoding?
> => false
> irb(main):005:0> shift_jis.encode!('UTF-8')
> => "\x93\xFA\x96{"
> irb(main):006:0> shift_jis.encoding
> => #<Encoding:UTF-8>
> irb(main):007:0> shift_jis.valid_encoding?
> => false
> ~~~

yaauie (commented):
I have left a couple questions about the UTF-8 value filtering.

spec/integration/inputs/kafka_spec.rb
```diff
@@ -91,6 +99,55 @@
         expect(event.get("[@metadata][kafka][timestamp]")).to be >= start
       end
     end
+
+    it "should show the right topic and group name in and kafka headers decorated kafka section" do
+      header = org.apache.kafka.common.header.internals.RecordHeader.new("name", "John".to_java_bytes)
```
yaauie, line 104: Since we are validating that UTF-8 can get through, we should include sequences from a variety of higher-order UTF-8 characters -- not just US-ASCII.

spec/integration/inputs/kafka_spec.rb
andsel: I used the emoji since they are not valid UTF-8, but your suggest is far better

lib/logstash/inputs/kafka.rb
andsel: So the correct way is to force the encoding in UTF8 and then check if the string is in a validig encoding?
```ruby
s = String.from_java_bytes(header.value)
s.force_encoding(Encoding::UTF-8)
s.valid_encoding?
```

neoaddix:
Hi, any chance on this one to be merged and release? I kind of really need it.

I'm also fine with doing manual install of the gem in the mean time (since we are still on test and not on prod) but I have zero experience with (j)ruby. If someone could build this gem for me than I can work with it on test until it is released

Thank you for any help!

lib/logstash/inputs/kafka.rb
```diff
@@ -292,14 +321,23 @@ def thread_runner(logstash_queue, consumer)
...
+              if @metadata_mode.include?(:record_props)
                 event.set("[@metadata][kafka][topic]", record.topic)
                 event.set("[@metadata][kafka][consumer_group]", @group_id)
                 event.set("[@metadata][kafka][partition]", record.partition)
                 event.set("[@metadata][kafka][offset]", record.offset)
                 event.set("[@metadata][kafka][key]", record.key)
                 event.set("[@metadata][kafka][timestamp]", record.timestamp)
               end
+              if @metadata_mode.include?(:headers)
+                for header in record.headers do
```
yaauie, line 333: nit: the ruby community [tends to avoid the `for X in Y` syntax](https://github.com/rubocop/ruby-style-guide#no-for-loops) because it leads to scoping surprises.

```suggestion
                record.headers.each do |header|
```

yaauie (approved):
One nitpick about using the `for` loop syntax, otherwise LGTM 👍🏼

PR #83, opened 2021-04-12
Docs: reflect clients update (since 10.6.0)

kares:
https://github.com/logstash-plugins/logstash-integration-kafka/commit/c0b6befad9492bd80cc419eb1582d715fb5f2189#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R67

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -2,8 +2,8 @@
 :plugin: kafka
 :type: input
 :default_codec: plain
-:kafka_client: 2.4
-:kafka_client_doc: 24
+:kafka_client: 2.5
+:kafka_client_doc: 25
 
 ///////////////////////////////////////////
 START - GENERATED VARIABLES, DO NOT EDIT!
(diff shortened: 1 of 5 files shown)


docs/input-kafka.asciidoc
```diff
@@ -2,8 +2,8 @@
 :plugin: kafka
 :type: input
 :default_codec: plain
-:kafka_client: 2.4
```
karenzone, line 5: Thanks for updating these vars. I'm happy to seem them being used. :-)

We used version `2.5.1` in the integration doc and version `2.5` for both input and output docs. We should be consistent.  Perhaps we should have your version checking in place before we commit to maintaining more granular versions (2.5.1 instead of 2.5).

karenzone (approved):
Builds cleanly and renders as expected. I left a comment inline for your consideration, but I trust your judgement on the resolution.  LGTM

docs/input-kafka.asciidoc
kares: The idea here is we need to have the precise version somewhere (to be able to know bug fixes etc).

For the input/output specific documentation it seemed okay to only have the minor version compatibility (the Kafka docs linked only uses minor version paths such as `/25/...`) but I am fine changing that if you prefer so.

I do have the version checking PR as a follow-up on this one.

karenzone: I'm fine with merging as is, and waiting for your version checking PR to go more granular on version info in input and output.  I have more changes I need to make anyway, so I can make the fixes after version checking is merged. Thanks!

PR #84, opened 2021-04-13
Test: check that Kafka version matches docs

kares:
a follow-up on https://github.com/logstash-plugins/logstash-integration-kafka/pull/83

The idea here being to have a failing test every time a Kafka clients upgrade is performed.
Also current output unit specs were assuming a running Kafka server - due improper mocking.


spec/check_docs_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    properties.load version.openStream
+    properties
+  end
+
+  it 'is sync-ed with Kafka client version' do
+    version = kafka_version_properties.get('version') # e.g. '2.5.1'
+
+    fails = docs_files.map do |file|
+      if line = File.readlines(file).find { |line| line.index(':kafka_client:') }
+        puts "found #{line.inspect} in #{file}" if $VERBOSE # e.g. ":kafka_client: 2.5\n"
```
kares, line 24: @karenzone we load the Kafka client version used by the plugin and than simply check the 3 docs files (*'index.asciidoc', 'input-kafka.asciidoc', 'output-kafka.asciidoc'*) for  a `:kafka_client: ...` whether it matches the `MAJOR.MINOR` or the full version `MAJOR.MINOR.PATCH`. isn't perfect but still useful to not fall out-of-sync for the next person doing a Kafka upgrade.

karenzone (commented):
This improvement is HUGE, @kares. Thank you for making this happen!

spec/unit/outputs/kafka_spec.rb
```diff
@@ -131,7 +134,7 @@
 
       it "should retry until successful" do
         count = 0
-
+        success = nil
         expect_any_instance_of(org.apache.kafka.clients.producer.KafkaProducer).to receive(:send)
               .exactly(sendcount).times
               .and_wrap_original do |m, *args|
```
yaauie, line 140: I _believe_ your changes in the block make it so we no longer need to wrap the original, since `m` is no longer touched.

```suggestion
              .exactly(sendcount).times do |*args|
```

spec/unit/outputs/kafka_spec.rb
```diff
@@ -141,15 +144,16 @@
 
             # Return a custom Future that will raise an exception to simulate a Kafka send() problem.
             future = java.util.concurrent.FutureTask.new { raise org.apache.kafka.common.errors.TimeoutException.new("Failed") }
-            future.run
-            future
           else
-            m.call(*args)
```
yaauie, line 147: I believe this orphans `m`, which means we don't need to be wrapping original?

spec/unit/outputs/kafka_spec.rb
```diff
@@ -141,15 +144,16 @@
...
+            success = args
```
yaauie, line 148: If we aren't validating the contents beyond truthiness, setting them to `args` (which will be an array containing the args sent to our producer) adds unnecessary details. We may as well set it to `true` directly.

spec/unit/outputs/kafka_spec.rb
```diff
@@ -141,15 +144,16 @@
...
+            future = java.util.concurrent.FutureTask.new { nil } # return no-op future
           end
+          future.tap { Thread.start { future.run } }
```
yaauie, line 151: it is not entirely clear to me why we need to run this future in another thread.

spec/unit/outputs/kafka_spec.rb
```diff
@@ -102,14 +105,14 @@
     before do
       count = 0
       expect_any_instance_of(org.apache.kafka.clients.producer.KafkaProducer).to receive(:send)
-        .exactly(1).times
-        .and_wrap_original do |m, *args|
+        .exactly(1).times.and_wrap_original do |m, *args|
```
yaauie, line 108: Similarly, we don't use `m` here so we don't need to wrap original.

```suggestion
        .exactly(1).times do |*args|
```

yaauie (commented):
I don't entirely understand what is going on in the "fixed" specs, especially with regard to needing to run futures in a separate thread, but I have added some notes on additional cleanup that is empowered by your changes.

The docs spec looks good to me.

spec/unit/outputs/kafka_spec.rb
kares: right, we do not want to be calling the original or do wrapping for these.

sorry was a bit of in a hurry (failing unit specs caught me by surprise) and did not take a second look.
should have reviewed the left-overs before submitting for review, I've cleaned up all unnecessary wrapping.
thanks for pointing these out :green_heart:

spec/unit/outputs/kafka_spec.rb
kares: previously all FutureTask run on the same thread in specs.
my understanding of the Kafka producer was that this isn't always the case (results are retrieved async-ly) so I though of adding some "real-world" behavior in to the mix where we get results/failures from other threads.

kares: giving it a few more days - otherwise I think the tests should be fine (will merge as is unless concerns are raised)

spec/check_docs_spec.rb
karenzone: Thanks for adding this check.

PR #87, opened 2021-05-10
Improve Kafka input error handling

robbavey:
This commit improves the error handling in the input side of this plugin by:

Adding begin...rescue blocks around the poll and commitSync methods in the polling thread to avoid handleable errors escaping out of the plugin.

Use Java threads instead of Ruby threads to enable the default uncaught exception handler to handle uncaught exceptions thrown in the polling thread, rather than deferring to the behavior determined by the value of `Thread.abort_on_exception`;  when this is set to true - as it currently is on `master` and `7.x` - such an error would cause Logstash execution to stop immediately across all pipelines. 

This commit does not current include any notion of backing off on errors.

### DIFF
--- a/lib/logstash/inputs/kafka.rb
+++ b/lib/logstash/inputs/kafka.rb
@@ -282,7 +282,8 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
   public
   def run(logstash_queue)
     @runner_consumers = consumer_threads.times.map { |i| subscribe(create_consumer("#{client_id}-#{i}")) }
-    @runner_threads = @runner_consumers.map { |consumer| thread_runner(logstash_queue, consumer) }
+    @runner_threads = @runner_consumers.map.with_index { |consumer, i| thread_runner(logstash_queue, consumer,
+                                                                                     "kafka-input-worker-#{client_id}-#{i}") }
     @runner_threads.each(&:start)
     @runner_threads.each(&:join)
   end # def run
@@ -303,8 +304,9 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
     consumer
   end
 
-  def thread_runner(logstash_queue, consumer)
+  def thread_runner(logstash_queue, consumer, name)
     java.lang.Thread.new do
+      LogStash::Util::set_thread_name(name)
       begin
         codec_instance = @codec.clone
         until stop?
--- a/lib/logstash/inputs/kafka.rb
+++ b/lib/logstash/inputs/kafka.rb
@@ -309,7 +309,7 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
         codec_instance = @codec.clone
         until stop?
           records = do_poll(consumer)
-            unless records.empty?
+          unless records.empty?
             records.each { |record| handle_record(record, codec_instance, logstash_queue) }
             maybe_commit_offset(consumer)
           end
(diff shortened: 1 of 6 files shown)


andsel (changes requested):
I like the refactoring done in the poller loop, really makes it more readable.
I've left a comment on the `nil` assignment that could be avoided.

lib/logstash/inputs/kafka.rb
```diff
@@ -296,62 +298,99 @@ def kafka_consumers
...
 
-  private
+  def subscribe(consumer)
+    @pattern.nil? ? consumer.subscribe(topics) : consumer.subscribe(@pattern)
+    consumer
+  end
+
   def thread_runner(logstash_queue, consumer)
-    Thread.new do
+    java.lang.Thread.new do
```
kares, line 307: just curious, are there other reasons for not using the (Ruby) `Thread` here, besides `abort_on_exception` ?

this is now going to act as a Java adopted thread (in Ruby land) and thus the runtime "tear-down" mechanics would slightly diverge - I do not think it will matter much but there's a difference on the thread not being managed by JRuby.

robbavey: Yeah, `abort_on_exception` handling was my main concern here. 

I would prefer to leave the handling of uncaught exceptions to the [uncaught exception handler](https://github.com/elastic/logstash/blob/c698aa224b44c7356fbfeb58ebb0c4a66d72dc71/logstash-core/src/main/java/org/logstash/Logstash.java#L96), rather than deferring to `abort_on_exception` - `Thread.abort_on_exception = true` will crash logstash on any 'escaped' exception, `Thread.abort_on_exception = false` will fail to shutdown when encountering fatal errors, such as `java.lang.OutOfMemoryError`.

kares: oh right, have assumed things might work if `abort_on_exception` would be disabled per thread here.
but that still means raising on the "main" thread instead of propagating to the Java thread handler.
have basically assumed [this part](https://github.com/jruby/jruby/blob/9.2.16.0/core/src/main/java/org/jruby/RubyThread.java#L1796-L1798) would allow `java.lang.Error` instances to also propagate through.

robbavey:
@kares, @andsel Ready for re-review - any more concerns on this PR?

andsel (approved):

lib/logstash/inputs/kafka.rb
```diff
@@ -309,7 +309,7 @@ def thread_runner(logstash_queue, consumer)
         codec_instance = @codec.clone
         until stop?
           records = do_poll(consumer)
-          unless records.nil? || records.count == 0
+            unless records.empty?
```
andsel, line 312: Minor, `unless` should be moved 1 indentation left

andsel (changes requested):

andsel (approved):

kares (approved):
LGTM, usage of `java.lang.Thread` is necessary due `Thread.start` keeping everything `java.lang.Throwable` around (some work on trying to [improve this for JRuby 9.3](https://github.com/jruby/jruby/pull/6679))

nit-pick: we could `LS::Util.set_thread_name`  naming threads `"kafka-worker_#1"` or smt

robbavey:
@kares - good idea, I've added the custom thread name

robbavey:
@kares Ready for final check

kares (approved):
:1st_place_medal:

PR #90, opened 2021-05-25
Improve wording on enable_auto_commit=false data safety

jsvd:
The current description of enable_auto_commit=false can lead to the interpretation that the commit happens right after data is fetched from a topic, while in truth that commit happens after that AND the data is written to the queue ([code here](https://github.com/logstash-plugins/logstash-integration-kafka/blob/master/lib/logstash/inputs/kafka.rb#L341-L345)).


karenzone (approved):
Nice clarification. Thank you. ❤️ 
Builds cleanly and LGTM

PR #91, opened 2021-05-31
Test: specify development dependency version

kares:
**digest-crc** since version 0.6.0 started using a [C-ext](https://github.com/postmodern/digest-crc/tree/v0.6.0/ext/digest)

this still works since the extension turns into a no-op on JRuby but the gem is still declared with:
```yaml
extensions:
- ext/digest/Rakefile
```
... which means the Rakefile will be triggered in an attempt to "build an extension"

being a test (development) dependency of ruby-kafka, 
forcing an olver ([extension-less](https://github.com/postmodern/digest-crc/tree/0.5.0)) version will effectively help the LS plugin install script to not run into issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 10.7.6
+  - Test: specify development dependency version [#91](https://github.com/logstash-plugins/logstash-integration-kafka/pull/91)
+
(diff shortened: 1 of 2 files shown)


andsel (approved):
LGTM

PR #94, opened 2021-06-23
Fix support for basic auth with the schema registry

robbavey:
This fixes an issue with supporting basic auth when using the schema registry

The plugin currently expects `schema_registry_secret` and `schema_registry_key` to
be set to authenticate with the schema registry. However, while this works during the
'register' phase, the authentication type `basic.auth.credentials.source` is not set
when creating the kafka consumer, leading to authentication failures when attempting
to deserialize the kafka payload.

This commit correctly sets the `basic.auth.credentials.source` to `USER_INFO` when
`schema_registry_key` and `schema_registry_secret` are set, and also enables the use
of a basic auth embedded in the `schema_registry_url` via `username:password@url`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 10.7.7
-  - Fix: Correct the settings to allow basic auth to work properly, either by setting 'schema_registry_key/secret` or embedding username/password in the
+  - Fix: Correct the settings to allow basic auth to work properly, either by setting `schema_registry_key/secret` or embedding username/password in the
     url [#94](https://github.com/logstash-plugins/logstash-integration-kafka/pull/94)
 
 ## 10.7.6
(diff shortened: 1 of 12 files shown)


robbavey:
This fixes #95

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 10.7.7
+  - Fix: Correct the settings to allow basic auth to work properly, either by setting 'schema_registry_key/secret` or embedding username/password in the
```
andsel, line 2:
```suggestion
  - Fix: Correct the settings to allow basic auth to work properly, either by setting `schema_registry_key/secret` or embedding username/password in the
```

spec/integration/inputs/kafka_spec.rb
```diff
@@ -292,37 +317,58 @@ def delete_remote_schema(schema_registry_client, subject_name)
       expect( subjects ).to be_empty
     end
   end
+end
+
+def cleanup_schema_registry
```
andsel, line 322: `cleanup_schema_registry` reminds me something that's cleaned on the Schema Registry service instead of a "shutdown Schema Registry" action

andsel (approved):
Very nice catch @robbavey.
I've left on a note on a bad backtick in CHANGELOG, plus a suggestion for rename the method responsible to shutdown the Schema Registry from test suites.

LGTM

PR #97, opened 2021-07-09
Add option to skip validation when dealing with schema registry

robbavey:
This PR adds a `validate_schema_registry` option to the plugin, allowing
users to bypass the schema registry validation while using a plugin. This is
a common issue when using authenticated schema registry, as the validation occurs
outside of the 'KafkaConsumer' code, and will require considerable effort to maintain
parity with that method.

This configuration option has 2 settings `auto` and `skip`. The default `auto` setting
will skip validation when schema registry auth is used that is known to fail, currently
when Kerberos authentication is used, but will try and validate in other cases. `skip`
will skip regardless of authentication method.

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -577,7 +577,7 @@ The schemas must follow a naming convention with the pattern <topic name>-value.
 Use either the Schema Registry config option or the
 <<plugins-{type}s-{plugin}-value_deserializer_class>> config option, but not both.
 
-[id="plugins-{type}s-{plugin}-validate_schema_registry"]
+[id="plugins-{type}s-{plugin}-schema_registry_validation"]
 ===== `schema_registry_validation`
 
   * Value can be either of: `auto`, `skip`
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -585,11 +585,9 @@ Use either the Schema Registry config option or the
 
 NOTE: Under most circumstances, the default setting of `auto` should not need to be changed.
 
-When using the schema registry, by default a validation process to check connectivity and
-the validity of the schema registry takes place during plugin registration. Under certain
-circumstances when trying to validate an authenticated schema registry, this process may
-fail, and cause the plugin to crash. This setting allows this validation to be skipped during
-registration.
+When using the schema registry, by default the plugin checks connectivity and validates the schema registry, during plugin registration, before events are processed.
+In some circumstances, this process may fail when it tries to validate an authenticated schema registry, causing the plugin to crash.
+This setting allows the plugin to skip validation during registration, which allows the plugin to continue and events to be processed. Note that an incorrectly configured schema registry will still stop the plugin from processing events.
 
 [id="plugins-{type}s-{plugin}-security_protocol"]
 ===== `security_protocol` 
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -128,6 +128,7 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-schema_registry_proxy>> |<<uri,uri>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_secret>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_url>> |<<uri,uri>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_validation>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-security_protocol>> |<<string,string>>, one of `["PLAINTEXT", "SSL", "SASL_PLAINTEXT", "SASL_SSL"]`|No
 | <<plugins-{type}s-{plugin}-send_buffer_bytes>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-session_timeout_ms>> |<<number,number>>|No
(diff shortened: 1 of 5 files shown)


docs/input-kafka.asciidoc
```diff
@@ -576,6 +577,20 @@ The schemas must follow a naming convention with the pattern <topic name>-value.
 Use either the Schema Registry config option or the
 <<plugins-{type}s-{plugin}-value_deserializer_class>> config option, but not both.
 
+[id="plugins-{type}s-{plugin}-validate_schema_registry"]
+===== `schema_registry_validation`
+
+  * Value can be either of: `auto`, `skip`
```
yaauie, line 583: 👍🏼 I like the `auto`/`skip` here. It conveys the complexity a lot better than a `true`/`false` would, and similarly doesn't convey promises that we will validate it (if it isn't set, or we can't validate) or that we won't ever do so (e.g., when we start using it and it doesn't provide what we need).

lib/logstash/plugin_mixins/common.rb
```diff
@@ -22,17 +22,30 @@ def setup_schema_registry_config
...
           @schema_registry_proxy_host, @schema_registry_proxy_port  = split_proxy_into_host_and_port(schema_registry_proxy)
           check_for_key_and_secret
-          check_for_schema_registry_connectivity_and_subjects
+          check_for_schema_registry_connectivity_and_subjects if schema_registry_validation?
         end
       end
 
+      def schema_registry_validation?
+        return false if schema_registry_validation.to_s == 'skip'
+        !using_kerberos?
```
yaauie, line 42: Nitpick:

I have a personal preference for what I call "bail conditions" being consistently formatted to make them more readable at a glance. Here, we have two conditions that would cause us to skip the validation:
 - `schema_registry_validation => skip`; OR
 - kerberos authentication

I think it would read a bit more clearly to call these out individually and return `true` only if we have not bailed yet, instead of relying on the `!using_kereberos?` for the final line.

```suggestion
        return false if schema_registry_validation.to_s == 'skip'
        return false if using_kerberos? # pre-validation doesn't support kerberos
        
        true
```

spec/unit/inputs/kafka_spec.rb
```diff
@@ -194,7 +199,52 @@
...
-    context "decorate_events" do
+    context 'when kerberos auth is used' do
+      ['SASL_SSL', 'SASL_PLAINTEXT'].each do |protocol|
+        context "with #{protocol}" do
+          ['auto', 'skip'].each do |vsr|
+            context "when validata_schema_registry is #{vsr}" do
+              let(:config) { base_config.merge({'security_protocol' => protocol,
+                                                'schema_registry_validation' => vsr})
+              }
+              it 'should skip verification' do
```
yaauie, line 210: nitpick, here and other instances: 

> ### "Should" in Example Docstrings
> 
> Do not write 'should' or 'should not' in the beginning of your example docstrings.
> The descriptions represent actual functionality, not what might be happening.
> Use the third person in the present tense.
> 
> ~~~ ruby
> # bad
> it 'should return the summary' do
>   # ...
> end
> 
> # good
> it 'returns the summary' do
>   # ...
> end
> ~~~
>
> -- https://github.com/rubocop/rspec-style-guide#should-in-example-docstrings

spec/unit/inputs/kafka_spec.rb
```diff
@@ -194,7 +199,52 @@
...
+        let(:config) { base_config.merge({'schema_registry_validation' => 'skip'})}
+        it 'should skip verification' do
+          expect(subject).not_to receive(:check_for_schema_registry_connectivity_and_subjects)
+          expect { subject.register }.not_to raise_error
+        end
+      end
+    end
+  end
+
+  context "decorate_events" do
```
yaauie, line 247: It looks like this example group got bumped up a level in the hierarchy, but the contents of the example group are still indented to match the older nesting.

yaauie (commented):
Functionality looks good. I've left a couple nitpicks.

yaauie (approved):
Code LGTM 👍🏼

docs/input-kafka.asciidoc
```diff
@@ -576,6 +577,20 @@ The schemas must follow a naming convention with the pattern <topic name>-value.
...
+  * Default value is `"auto"`
+
+NOTE: Under most circumstances, the default setting of `auto` should not need to be changed.
+
+When using the schema registry, by default a validation process to check connectivity and
+the validity of the schema registry takes place during plugin registration. Under certain
+circumstances when trying to validate an authenticated schema registry, this process may
+fail, and cause the plugin to crash. This setting allows this validation to be skipped during
+registration.
```
karenzone, line 592:
```suggestion
When using the schema registry, the plugin checks connectivity and validates the schema registry by default. 
In some circumstances, this process may fail when it tries to validate an authenticated schema registry, causing the plugin to crash.
This setting allows the plugin to skip validation during registration.
```

karenzone: Suggested rewording for consideration.

karenzone (changes requested):
Headings don't match between the table and description, and would crash the docs build. Pick your favorite. 

I left suggestion wording inline for your consideration. We do some of our best work when we iterate, so your suggestions to my suggestions are welcome (as always). :-)

docs/input-kafka.asciidoc
robbavey:
```suggestion
When using the schema registry, by default the plugin checks connectivity and validates the schema registry, during plugin registration, before events are processed.
In some circumstances, this process may fail when it tries to validate an authenticated schema registry, causing the plugin to crash.
This setting allows the plugin to skip validation during registration, which allows the plugin to continue and events to be processed. Note that an incorrectly configured schema registry will still stop the plugin from processing events.
```

robbavey: @karenzone How about this?

karenzone: I'm a fan! Let's go with it.

karenzone: The missing reference problem has been resolved. Wasn't sure it you wanted me to commit your suggestion and then approve, or if you had a different workflow in mind. It should be an easy signoff whenever you're ready for final review.

robbavey:
@karenzone I've made the changes, ready for another round

karenzone (approved):
LGTM

Issue #20, opened 2020-03-13
[Docs] Update docs to reflect behaviour with Kafka 2.x


robbavey:
Currently, in the docs we advise that `The value of the configuration request_timeout_ms must always be larger than max_poll_interval_ms`.

According to [this stack overflow article](https://stackoverflow.com/questions/54822434/what-is-negative-effects-of-setting-max-poll-interval-ms-larger-than-request-tim#:~:text=According%20to%20Kafka%20documentation%3B,(5%20minutes%20by%20default).), it looks like this advice is no longer accurate when dealing with Kafka 2.0+  

I believe the advice originally came from [the 10.1.0 upgrade notes](https://kafka.apache.org/documentation/#upgrade_1010_notable):

```
The new Java Consumer now supports heartbeating from a background thread. There is a new configuration max.poll.interval.ms which controls the maximum time between poll invocations before the consumer will proactively leave the group (5 minutes by default). **The value of the configuration request.timeout.ms must always be larger than max.poll.interval.ms** because this is the maximum time that a JoinGroup request can block on the server while the consumer is rebalancing, so we have changed its default value to just above 5 minutes. Finally, the default value of session.timeout.ms has been adjusted down to 10 seconds, and the default value of max.poll.records has been changed to 500.
```

But that now appears to be outdated as of [the 2.0.0 upgrade notes](https://kafka.apache.org/documentation/#upgrade_200_notable):

```
Also as part of KIP-266, the default value of request.timeout.ms has been changed to 30 seconds. The previous value was a little higher than 5 minutes to account for maximum time that a rebalance would take. Now we treat the JoinGroup request in the rebalance as a special case and use a value derived from max.poll.interval.ms for the request timeout. All other request types use the timeout defined by request.timeout.ms
```

PR #99, opened 2021-07-13
Doc: Remove outdated recommendation for settings

karenzone:
Remove a setting recommendation that is no longer accurate for Kafka 2.0+. 

Fixes: #20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 10.8.1
+  - [DOC] Removed a setting recommendation that is no longer applicable for Kafka 2.0+ [#99](https://github.com/logstash-plugins/logstash-integration-kafka/pull/99)
+
 ## 10.8.0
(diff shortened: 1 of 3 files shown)


karenzone:
@robbavey @psanz-estc What else is needed? Any further guidance we can provide?

robbavey (approved):
I'm good with this - it removes confusing and out of date information from the docs. I think there's a broader issue around our use of defaults for this plugin, how they relate to the defaults from the kafka library and the appropriateness of those defaults for Logstash, but that is a bigger task than this one and removing stale information is important.

tl;dr: LGTM

karenzone:
100% agreed. We haven't solved the broader settings/document issue because it's not easy to solve. At least this is a tiny step in the right direction.

karenzone:
Published at v10.8.1: https://rubygems.org/gems/logstash-integration-kafka/versions/10.8.1-java

PR #103, opened 2021-10-13
[doc] clarify Kafka client does not support proxy

kaisecheng:


docs/input-kafka.asciidoc
```diff
@@ -183,6 +183,8 @@ for the initial connection to discover the full cluster membership (which may ch
 so this list need not contain the full set of servers (you may want more than one, though, in
 case a server is down).
 
+NOTE: Kafka client does not support proxy setup. Only schema registry supports proxy.
```
roaksoax: @kaisecheng I think we need improved wording, e.g.:

This plugin only supports using a proxy when communicating to the Schema Registry <this should link to the setting>. This plugin does not support using a proxy when communicating to Kafka. This is because of a limitation in the Kafka Client, which this plugin uses. For more information, please refer to the upstream documentation <this should link to the documentation>.

@kaisecheng also, I'm, guessing some of this wording / note is required for the Kafka output.

@karenzone I think this should be in the special format to render notes, right?

roaksoax (changes requested):

docs/input-kafka.asciidoc
karenzone: You're right! We do need to use special formatting to render notes correctly.  Kaise nailed it in the original commit.  :-)

docs/input-kafka.asciidoc
```diff
@@ -42,6 +42,10 @@ This input supports connecting to Kafka over:
 
 By default security is disabled but can be turned on as needed.
 
+Due to the limitation of https://javadoc.io/doc/org.apache.kafka/kafka-clients/2.5.1/org/apache/kafka/clients/consumer/ConsumerConfig.html[Kafka client],
+this plugin does not support using a proxy when communicating to Kafka broker.
+This plugin only supports using a proxy when communicating to the Schema Registry, see <<plugins-{type}s-{plugin}-schema_registry_proxy>>
```
karenzone, line 47:
```suggestion
[NOTE]
=======
This plugin does not support using a proxy when communicating to the Kafka broker.  

This plugin does support using a proxy when communicating to the Schema Registry using the <<plugins-{type}s-{plugin}-schema_registry_proxy>> option.
=======
```

karenzone: Additional thoughts for discussion (@roaksoax @kaisecheng): 

- @kaisecheng: The NOTE format you used originally is correct. I'm proposing some advanced formatting for this particular situation. The takeaway: The formatting I'm proposing here isn't something we want to adopt except in special situations. 
- The link to the kafka docs doesn't appear to have anything specific to proxy support.  Unless we can point to something more specific or explain the reason we're linking to the kafka docs, I propose that we don't include the link. There are already other links to the kafka client in the docs and it's important not to use external links unless they are highly relevant. 
- I'm hesitant to say that another vendor has limitations in their product. Let's discuss the reasons.

karenzone: The NOTE format suggested above will render like this: 
Screen Shot 2021-10-29 at 6 38 18 PM

karenzone (commented):
Left suggestions and comments for discussion.

docs/input-kafka.asciidoc
karenzone: REMINDER: Earlier comments indicated that we might need to include this content for output-kafka. Have we determined if that is or isn't the needed?

kaisecheng: @karenzone I am with you hesitating to link the java doc to show Kafka client has a limitation on supporting proxy. 
I cannot find any doc on Kafka side explicitly saying they don't support it, but we found from the code level that they aren't. I feel more comfortable removing the link.

kaisecheng: Good point! I should update the output doc as well 👍

karenzone (approved):
LGTM

PR #106, opened 2021-12-02
Refactor: leverage codec when using schema registry

kares:
The motivation here is to be able to leverage the proper codec in all scenarios the plugin operates.

The default behavior for the Kafka input is to use the plain codec and decode the record payload.
With the `schema_registry_url` the codec is still used but the payload is than parsed again as json.

The PR changes the behavior with some clever hooks into `Plugin#initialize` to change the default codec:
- when user specifies `schema_registry_url` default to using the json codec
- when `schema_registry_url` is not specified default to the plain codec
- when the user specified a `codec` option use that regardless of `schema_registry_url`

NOTE: PR's changes cause a slight regression, previously the plugin worked fine using `schema_registry_url` when a codec (e.g. `codec => plain`) was also specified - since json parsing was done manually regardless of the codec.

---

The changes are sufficient given that `ecs_compatibility` is either local or a global setting -> JSON codec default in case of the schema registry will already warn users on a missing target e.g. 

```
[2021-12-06T12:55:15,785][INFO ][org.reflections.Reflections] Reflections took 116 ms to scan 1 urls, producing 119 keys and 416 values 
[2021-12-06T12:55:16,222][INFO ][logstash.codecs.json     ] ECS compatibility is enabled but `target` option was not specified. This may cause fields to be set at the top-level of the event where they are likely to clash with the Elastic Common Schema. It is recommended to set the `target` option to avoid potential schema conflicts (if your data is ECS compliant or non-conflicting, feel free to ignore this message)
[2021-12-06T12:55:16,646][INFO ][logstash.javapipeline    ] Pipeline `main` is configured with `pipeline.ecs_compatibility: v8` setting. All plugins in this pipeline will default to `ecs_compatibility => v8` unless explicitly configured otherwise.
[2021-12-06T12:55:16,786][INFO ][logstash.javapipeline    ][main] Starting pipeline {:pipeline_id=>"main", "pipeline.workers"=>16, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>50, "pipeline.max_inflight"=>2000, "pipeline.sources"=>["config string"], :thread=>"#<Thread:0x7a8ed04a run>"}

```

### DIFF
--- a/lib/logstash/inputs/kafka.rb
+++ b/lib/logstash/inputs/kafka.rb
@@ -63,7 +63,10 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
 
   config_name 'kafka'
 
-  default :codec, 'plain'
+  # default :codec, 'plain' or 'json' depending whether schema registry is used
+  #
+  # @override LogStash::Inputs::Base - removing the `:default => :plain`
+  config :codec, :validate => :codec # default: plain or json (when using schema_registry)
 
   # The frequency in milliseconds that the consumer offsets are committed to Kafka.
   config :auto_commit_interval_ms, :validate => :number, :default => 5000 # Kafka default
@@ -249,6 +252,14 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
 
   attr_reader :metadata_mode
 
+  def initialize(params = {})
+    unless params.key?('codec')
+      params['codec'] = params.key?('schema_registry_url') ? 'json' : 'plain'
+    end
+
+    super(params)
+  end
+
   public
   def register
     @runner_threads = []
@@ -341,19 +352,8 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
   def handle_record(record, codec_instance, queue)
     codec_instance.decode(record.value.to_s) do |event|
       decorate(event)
-      maybe_apply_schema(event, record)
       maybe_set_metadata(event, record)
       queue << event
-    end
-  end
-
-  def maybe_apply_schema(event, record)
-    if schema_registry_url
-      json = LogStash::Json.load(record.value.to_s)
-      json.each do |k, v|
-        event.set(k, v)
-      end
-      event.remove("message")
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/logstash/inputs/kafka.rb
```diff
@@ -249,6 +252,14 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
 
   attr_reader :metadata_mode
 
+  def initialize(params = {})
+    unless params.key?('codec')
+      params['codec'] = params.key?('schema_registry_url') ? 'json' : 'plain'
+    end
+
+    super(params)
```
yaauie, line 260: I have thought about it a lot, and this is certainly the best place to perform the injection. We get fully-contextualized codecs as a result.

This has the minor side-effects of (1) putting `codec` in the `original_params` when it wasn't specified and (2) mutating the hash that was sent. 

I'm not sure it's worth the effort, but the following doesn't have those side-effects.

```suggestion
    return super if params.key?('codec')
    
    params_with_codec = params.merge('codec' => (params.key?('schema_registry_url') ? 'json' : 'plain')
    super(params_with_codec)
    @original_params.delete('codec') # we were never here...
```

yaauie (commented):
conceptually I _think_ this will work, because we are passing `record.value.to_s` to the codec anyway, but I don't fully understand how the presence of the `schema_registry_url`, which enforces the value deserializer class to be avro, will also ensure that the contents of `record.value.to_s` will be json.

I've left one comment about cleaning up side-effects. I'm willing to LGTM the current iteration if you don't think that the cleanup is useful, but I'm hoping the side-effects won't surprise us.

lib/logstash/inputs/kafka.rb
kares: both the `super` (from [`config_init`](https://github.com/elastic/logstash/blob/v7.15.2/logstash-core/lib/logstash/config/mixin.rb#L94)) and super's super ([`Plugin#initialize`](https://github.com/elastic/logstash/blob/v7.15.2/logstash-core/lib/logstash/plugin.rb#L79)) change the input `params`.

the work here was inspired by existing behavior -> fill in the `params['codec']` default before `config_init`

this is not ideal but the contract is broken already in core in multiple places, at this point I would be more concerned about breaking smt that relies on the existing core behavior of passing in `params` and having all defaults filled in in the same Hash instance.

yaauie (approved):
LGTM :+1:

PR #110, opened 2022-01-14
Fix(docs): Discourage should be about changing defaults, not using them

joostdecock:
While I can't read minds, I am fairly certain that what is *discouraged* here is to change this default setting, not to use the default setting.

I think it's fair to assume that no product would ship with default settings that are discouraged. So I submit this fix to make the documentation match the intention.


jsvd (approved):
LGTM

PR #119, opened 2022-05-09
Fix CHANGELOG.md link to PR

karenzone:
Updates changelog link to point to PR


karenzone:
@jsvd The current link resolves, but I thought we might want to link to the PR instead.  If not, we can close without merge.

Issue #116, opened 2022-04-07
Document how topics pattern regex happens at kafka clients instead of broker


jsvd:
If a user sets `topics_pattern => /regex/` what happens is that the kafka client used by this integration requests a full list of topics from the broken, then applies the regex locally.

This has two potential consequences:
1. if the kafka broker has a lot of topics this operation can be very slow, especially if there are also many consumers
2. if kafka broker has ACL enabled and the user only has WRITE/DESCRIBE permissions on the topics that match the regex, then the user will be surprised that authorization exceptions will be returned for topics that don't match the pattern (because the client request the full list).

This is known issue https://issues.apache.org/jira/browse/KAFKA-10717 and we should note this in the documentation to avoid surprises.

PR #122, opened 2022-07-05
Describe potenatial problems in `topic_pattern` setting

andsel:
Added a couple of notes to make the user aware of some performance and access permission problems that could happen using the `topic_pattern` setting.

Fixes: #116

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -699,8 +699,14 @@ A list of topics to subscribe to, defaults to ["logstash"].
   * Value type is <<string,string>>
   * There is no default value for this setting.
 
-A topic regex pattern to subscribe to. 
-The topics configuration will be ignored when using this configuration.
+A topic regex pattern to subscribe to; retrieve the full list from the broker and filters applying the pattern locally.
+The <<plugins-{type}s-{plugin}-topics>> configuration will be ignored when using this configuration.
+
+NOTE: When used with brokers with a lot of topics, this operation, could be very slow especially if there are a lof of consumers.
+
+NOTE: When the broker has some ACL rules, configured with WRITE or DESCRIBE permissions, there are some unexpected access permissions
+on topics that doesn't match the regex pattern. This is due to the fact that the full list of topics has to be retrieved.
+
 
 [id="plugins-{type}s-{plugin}-value_deserializer_class"]
 ===== `value_deserializer_class` 


docs/input-kafka.asciidoc
```diff
@@ -699,8 +699,14 @@ A list of topics to subscribe to, defaults to ["logstash"].
   * Value type is <<string,string>>
   * There is no default value for this setting.
 
-A topic regex pattern to subscribe to. 
-The topics configuration will be ignored when using this configuration.
+A topic regex pattern to subscribe to; retrieve the full list from the broker and filters applying the pattern locally.
+The <<plugins-{type}s-{plugin}-topics>> configuration will be ignored when using this configuration.
+
+NOTE: When used with brokers with a lot of topics, this operation, could be very slow especially if there are a lof of consumers.
```
jsvd, line 705:
```suggestion
A topic regular expression pattern to subscribe to.

Filtering by a regular expression is done by retrieving the full list of topic names from the broker and applying the pattern locally. When used with brokers with a lot of topics this operation could be very slow, especially if there are a lot of consumers.
```

docs/input-kafka.asciidoc
```diff
@@ -699,8 +699,14 @@ A list of topics to subscribe to, defaults to ["logstash"].
...
+
+NOTE: When the broker has some ACL rules, configured with WRITE or DESCRIBE permissions, there are some unexpected access permissions
+on topics that doesn't match the regex pattern. This is due to the fact that the full list of topics has to be retrieved.
```
jsvd, line 708: Can we make it clearer to the user what happens in this case? Logstash will retrieve the full list of topics, and some will have ACLs. Does it mean that topics Logstash can't access are skipped? is there an error shown? does the pipeline crash?

andsel: :+1:  the subscription happens has expected but the broker logs line as 
```
Principal = User:<Logstash username> is Denied Operation = Describe from host ...
```

docs/input-kafka.asciidoc
```diff
@@ -699,8 +699,13 @@ A list of topics to subscribe to, defaults to ["logstash"].
...
   * There is no default value for this setting.
 
-A topic regex pattern to subscribe to. 
-The topics configuration will be ignored when using this configuration.
+A topic regular expression pattern to subscribe to.
+
+Filtering by a regular expression is done by retrieving the full list of topic names from the broker and applying the pattern locally. When used with brokers with a lot of topics this operation could be very slow, especially if there are a lot of consumers.
+
+NOTE:  When the broker has some topics configured with ACL rules and they miss the DESCRIBE permission, then the subscription
+happens as expected but on broker side is logged the fact that it denied to describe some topics from the Logstash's user.
```
jsvd, line 707:
```suggestion
happens but on the broker side it is logged that the subscription of some topics was denied to the configured user.
```

PR #123, opened 2022-07-14
Update kafka version for integration tests

robbavey:
The version of Kafka prior to this commit included a version of zookeeper that
was incompatible with JDK17. This commit updates the version of Kafka to a more
modern version that includes a JDK17 compatible zookeeper


andsel (approved):
LGTM

PR #135, opened 2023-01-13
Implement the `static membership` feature to Kafka input

### LINKED ISSUE #134: Add `static membership` Kafka feature to clients
The `static membership` concept is proposed and explained in [KIP-345](https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances).

When a client of a consumer group reconnects to a cluster it triggers a re-balance, to assign partition to consumers and this consumer is assigned an id named `member.id` which is randomly generated.
If a client is member of a consumer groups, and reconnects frequently is mean it triggers frequent re-balances.
To avoid this re-assignment of consumer to partitions during re-balancing, a static member.id could be passed by the consumer, to uniquely identify it inside the consumer group and in stored on both sides (broker and consumer). In this way the static `group.instance.id`  avoids re-balances in case of continuous disconnections and re-connections.
Also with Static Membership however, re-balances could happen when a new member joins the group or a leader re-joins, for a full list of cases where re-balances happen also with static membership, please check [KIP-345](https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances)

andsel:
## Release notes

Added config `group_instance_id` to use the Kafka's consumer static membership feature

## What does this PR do?

Static membership is reflected in the Kafka property `group.instance.id`, which has to be a unique identifier of the consumer instance provided by end user.
If `consumer_threads` settings is 1 the value is passed directly down to the Kafka's consumer configuration, but if threads count is more than 1, as per [KIP-345](https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances) it would clash, so in that case, a postfix `-<thread-index>` is added.

## Why is it important/What is the impact to the user?

The [static membership](https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances) feature offered by Kafka client's consumer is intended to bind a consumer to a partition, this is needed in cases where the cost of state replication between consumers during a rebalance, is high. This PR exposes the feature, which is optional, to Logstash's users.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- ~~[ ] I have made corresponding change to the default configuration files (and/or docker env variables)~~
- [x] I have added tests that prove my fix is effective or that my feature works

## Author's Checklist

- [x] test locally on rel Kafka with 1 thread and more threads
- [x] test locally with another client that kicks off the Kafka input plugin
- [x] obtain a  docs review.

### Runs a local Kafka cluster
Fro the clone of this repository launch the test Kafka script
```bash
./kafka_test_setup.sh
```

### Connect a producer
- start the producer with
```bash
bin/kafka-console-producer.sh --bootstrap-server localhost:9092 --topic "logstash_integration_static_membership_topic"
```

### Setup Logstash Kafka input & run 
- Adds the plugin definition in `Gemfile`
```
gem "logstash-integration-kafka", :path => "/path/to/logstash_plugins/logstash-integration-kafka"
```
- Install the plugin in development mode
```bash
bin/logstash-plugin install --no-verify
```
- edit the pipeline `test_pipeline.conf` file
```
input {
  kafka {
    bootstrap_servers => ["localhost:9092"]
    topics => ["logstash_integration_static_membership_topic"]
    group_id => "logstash"
    consumer_threads => 1

    group_instance_id => "test_static_group_id"
  }
}


output {
  stdout {
    codec => rubydebug
  }
}
```

### Verify Logstash is receiving data
From the producer's console send some data and verify on the Logstash console the message is received.

### Verify another consumer kicks off Logstash
Start another consumer with same `group.instance.id` in same consumer group
- In the folder `build/kafka the file `client_config.properties` with content
```
group.instance.id=test_static_group_id
```
```bash
bin/kafka-console-consumer.sh --bootstrap-server localhost:9092 --topic "logstash_integration_static_membership_topic" --from-beginning --group logstash --consumer.config "${PWD}/client_config.properties"
```

### Test with multiple threads
- Edit the pipeline to have `consumer_threads` > 1 and verify that another consumer doesn't kicks off because the `group.instance.id` has now been suffixed with `-n` for each Logstash Kafka consumer threads.

## Related issues

- Closes #134

## Use cases

As a Kafka uses that want to avoid that consumers spend a lot of time during rebalances, specially when they have heavy state, I want to be able to assign a "static membership" id to every consumer.

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -359,7 +359,9 @@ This mitigates the cases where the service state is heavy and rebalance of one t
 A to B means huge amount of data transfer. Client that goes offline and online frequently, if using this settings
 avoid frequent and heavy rebalances.
 
-NOTE: In case another client connects with same `group.instance.id` value then the oldest is kicked off.
+NOTE: It has to be unique across all the clients connected to same topic, in case another client connects
+with same `group.instance.id` value then the oldest is kicked off.
+Generally could be something like a hostname, an IP or anything that uniquely identifies the client application.
 
 NOTE: In cases when multiple threads are configured, `consumer_threads` is greater than one, a suffix is appended to
 the `group_instance_id` to avoid collisions.
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -113,6 +113,7 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-fetch_max_wait_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-fetch_min_bytes>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-group_id>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-group_instance_id>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-heartbeat_interval_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-isolation_level>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-jaas_path>> |a valid filesystem path|No
@@ -343,6 +344,25 @@ Logstash instances with the same `group_id`.
 NOTE: In cases when multiple inputs are being used in a single pipeline, reading from different topics,
 it's essential to set a different `group_id => ...` for each input. Setting a unique `client_id => ...`
 is also recommended.
+
+[id="plugins-{type}s-{plugin}-group_instance_id"]
+===== `group_instance_id`
+
+* Value type is <<string,string>>
+* Optional, there is no default value for this setting.
+
+The static membership identifier for this consumer. Static membership feature was introduced in
+https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances[KIP-345],
+available under Kafka property `group.instance.id` and aims to avoid rebalances in contexts where a lot of data
+has to be handed off after a consumer goes offline.
+This mitigates the cases where the service state is heavy and rebalance of one topic partition from instance
+A to B means huge amount of data transfer. Client that goes offline and online frequently, if using this settings
+avoid frequent and heavy rebalances.
+
+NOTE: In case another client connects with same `group.instance.id` value then the oldest is kicked off.
+
+NOTE: In cases when multiple threads are configured, `consumer_threads` is greater than one, a suffix is appended to
+the `group_instance_id` to avoid collisions.
 
 [id="plugins-{type}s-{plugin}-heartbeat_interval_ms"]
 ===== `heartbeat_interval_ms` 
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -360,8 +360,8 @@ This feature mitigates cases where the service state is heavy and the rebalance 
 A to B would cause a huge amount of data to be transferred.
 A client that goes offline/online frequently can avoid frequent and heavy rebalances by using this option.
 
-NOTE: It has to be unique across all the clients belonging to the same <<plugins-{type}s-{plugin}-group_id>>, in case another client connects
-with same `group.instance.id` value then the oldest is kicked off.
+NOTE: The `group_instance_id` setting must be unique across all the clients belonging to the same <<plugins-{type}s-{plugin}-group_id>>.
+Otherwise, another client connecting with same `group.instance.id` value would cause the oldest instance to be disconnected.
 You can set this value to use information such as a hostname, an IP, or anything that uniquely identifies the client application.
 
 NOTE: In cases when multiple threads are configured and `consumer_threads` is greater than one, a suffix is appended to
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -359,7 +359,7 @@ This mitigates the cases where the service state is heavy and rebalance of one t
 A to B means huge amount of data transfer. Client that goes offline and online frequently, if using this settings
 avoid frequent and heavy rebalances.
 
-NOTE: It has to be unique across all the clients connected to same topic, in case another client connects
+NOTE: It has to be unique across all the clients belonging to the same <<plugins-{type}s-{plugin}-group_id>>, in case another client connects
 with same `group.instance.id` value then the oldest is kicked off.
 Generally could be something like a hostname, an IP or anything that uniquely identifies the client application.
 
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -349,21 +349,22 @@ is also recommended.
 ===== `group_instance_id`
 
 * Value type is <<string,string>>
-* Optional, there is no default value for this setting.
+* There is no default value for this setting.
 
-The static membership identifier for this consumer. Static membership feature was introduced in
+The static membership identifier for this Logstash Kafka consumer. Static membership feature was introduced in
 https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances[KIP-345],
-available under Kafka property `group.instance.id` and aims to avoid rebalances in contexts where a lot of data
-has to be handed off after a consumer goes offline.
-This mitigates the cases where the service state is heavy and rebalance of one topic partition from instance
-A to B means huge amount of data transfer. Client that goes offline and online frequently, if using this settings
-avoid frequent and heavy rebalances.
+available under Kafka property `group.instance.id`.
+Its purpose is to avoid rebalances in situations in which a lot of data
+has to be forwarded after a consumer goes offline.
+This feature mitigates cases where the service state is heavy and the rebalance of one topic partition from instance
+A to B would cause a huge amount of data to be transferred.
+A client that goes offline/online frequently can avoid frequent and heavy rebalances by using this option.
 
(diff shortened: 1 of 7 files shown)


lib/logstash/inputs/kafka.rb
```diff
@@ -287,7 +292,10 @@ def extract_metadata_level(decorate_events_setting)
 
   public
   def run(logstash_queue)
-    @runner_consumers = consumer_threads.times.map { |i| subscribe(create_consumer("#{client_id}-#{i}")) }
+    @runner_consumers = consumer_threads.times.map do |i|
+      thread_group_instance_id = consumer_threads > 1 && group_instance_id ? "#{group_instance_id}-#{i}" : group_instance_id
```
jsvd, line 296: Any thoughts on always adding "i" to the name? this has an upside and downside..downside is that we're not using the exact name the user set when threads = 1, the upside is that naming is consistent regardless the number of threads, and also consistent with client_id naming.

andsel: I had the same thought, but given that `consumer_threads` is 1 by default, I thought not adding it, was the minimal surprising behavior for users.

lib/logstash/inputs/kafka.rb
```diff
@@ -335,6 +343,9 @@ def do_poll(consumer)
     rescue org.apache.kafka.common.errors.WakeupException => e
       logger.debug("Wake up from poll", :kafka_error_message => e)
       raise e unless stop?
+    rescue org.apache.kafka.common.errors.FencedInstanceIdException => e
+      logger.error("Another consumer with same group.instance.id has connected")
```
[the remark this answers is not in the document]
andsel: I think we could just add the `:group_instance_id` but in case of `consumer_threads` is > 1 that would be inaccurate and maybe it's better to add the exception message. Other, we should propagate the effective  `group_instance_id` down to the `#do_poll` method.
WDYT?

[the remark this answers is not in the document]
jsvd: you can abstract away the value with a method, something like this:

```ruby
def group_instance_id_w_threads
  return unless group_instance_id
  consumer_threads > 1 ? "#{group_instance_id}-#{i}" : group_instance_id
end
```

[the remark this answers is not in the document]
andsel: Sorry, I don't understand how it solves the problem of propagate the "number of thread" down to the `#do_poll`.

The message of the caught `org.apache.kafka.common.errors.FencedInstanceIdException` exception is something like hte following:
```
Get fenced exception for group.instance.id test_group_id, current member.id is test_group_id-539e9f45-f91f-4317-93d5-20cd78dddbc6
```

so it already contains the effective `group.instance.id ` used by the consumer instance.
We could print it, but in that case we bind to the exception, and if in a future version of the Kafka library that information is removed we also lost it. I think it's unlikely that it happens, but we would rely on the client exception to provide information that's valuable for the customer.

spec/integration/inputs/kafka_spec.rb
```diff
@@ -185,10 +186,109 @@ def send_message(record)
...
+        wait_kafka_input_is_ready("logstash_integration_static_membership_topic", queue)
+        saboteur_kafka_consumer = create_consumer_and_start_consuming("test_static_group_id")
+        saboteur_kafka_consumer.run # ask to be scheduled
+        saboteur_kafka_consumer.join
+
+        expect(saboteur_kafka_consumer.value).to eq("saboteur exited")
+      ensure
+        input_worker.join(30_000)
+      end
+    end
```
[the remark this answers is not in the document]
andsel: Nice suggestion thanks, integrated with commit 97f35b3344a175a6bb365185b9d311d0ff81ed8a.

lib/logstash/inputs/kafka.rb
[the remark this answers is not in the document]
jsvd: `do_poll` has access to `consumer_threads` since it's a setting. but if the exception message already contains the id then nevermind :)

lib/logstash/inputs/kafka.rb
```diff
@@ -407,6 +418,10 @@ def create_consumer(client_id)
       props.put(kafka::FETCH_MAX_WAIT_MS_CONFIG, fetch_max_wait_ms.to_s) unless fetch_max_wait_ms.nil?
       props.put(kafka::FETCH_MIN_BYTES_CONFIG, fetch_min_bytes.to_s) unless fetch_min_bytes.nil?
       props.put(kafka::GROUP_ID_CONFIG, group_id)
+      if group_instance_id
+        # group_instance_id in case of nil can't be set
+        props.put(kafka::GROUP_INSTANCE_ID_CONFIG, group_instance_id)
+      end
```
jsvd, line 424:
```suggestion
      props.put(kafka::GROUP_INSTANCE_ID_CONFIG, group_instance_id) unless group_instance_id.nil?
```

jsvd (approved):
LGTM

lib/logstash/inputs/kafka.rb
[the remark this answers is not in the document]
andsel: Using the exception message the error log becomes:
```
"Another consumer with same group.instance.id has connected", {:original_error_message=>"The broker rejected this static consumer since another consumer with the same group.instance.id has registered with a different member.id."}
```
which helps  more than before in trouble shooting possible problems.

docs/input-kafka.asciidoc
```diff
@@ -344,6 +345,27 @@ NOTE: In cases when multiple inputs are being used in a single pipeline, reading
...
+[id="plugins-{type}s-{plugin}-group_instance_id"]
+===== `group_instance_id`
+
+* Value type is <<string,string>>
+* Optional, there is no default value for this setting.
+
+The static membership identifier for this consumer. Static membership feature was introduced in
+https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances[KIP-345],
+available under Kafka property `group.instance.id` and aims to avoid rebalances in contexts where a lot of data
+has to be handed off after a consumer goes offline.
```
karenzone, line 357:
```suggestion
available under Kafka property `group.instance.id`.
Its purpose is to avoid rebalances in situations in which a lot of data
has to be forwarded after a consumer goes offline.
```

docs/input-kafka.asciidoc
```diff
@@ -344,6 +345,27 @@ NOTE: In cases when multiple inputs are being used in a single pipeline, reading
...
+This mitigates the cases where the service state is heavy and rebalance of one topic partition from instance
+A to B means huge amount of data transfer. Client that goes offline and online frequently, if using this settings
+avoid frequent and heavy rebalances.
```
karenzone, line 360:
```suggestion
This feature mitigates cases where the service state is heavy and the rebalance of one topic partition from instance
A to B would cause a huge amount of data to be transferred.
A client that goes offline/online frequently can avoid frequent and heavy rebalances by using this option.
```

docs/input-kafka.asciidoc
```diff
@@ -344,6 +345,27 @@ NOTE: In cases when multiple inputs are being used in a single pipeline, reading
...
+
+NOTE: It has to be unique across all the clients belonging to the same <<plugins-{type}s-{plugin}-group_id>>, in case another client connects
+with same `group.instance.id` value then the oldest is kicked off.
```
karenzone, line 363:
```suggestion
NOTE: The `group_instance_id` setting must be unique across all the clients belonging to the same <<plugins-{type}s-{plugin}-group_id>>. Otherwise, another client connecting
with same `group.instance.id` value would cause the oldest instance to be removed.
```

karenzone: I'm not sure I got all the details right.  I'm trying to add precise language by restating the name of the option rather than just saying "it".  
Also, "kicked off" is ambiguous in English.  It can mean to start something (like kicking off a process" or deleting/removing something.  I'm guessing you mean the latter.

docs/input-kafka.asciidoc
```diff
@@ -344,6 +345,27 @@ NOTE: In cases when multiple inputs are being used in a single pipeline, reading
...
+Generally could be something like a hostname, an IP or anything that uniquely identifies the client application.
```
karenzone, line 364:
```suggestion
You can set this value to use information such as a hostname, an IP, or anything that uniquely identifies the client application.
```

docs/input-kafka.asciidoc
```diff
@@ -344,6 +345,27 @@ NOTE: In cases when multiple inputs are being used in a single pipeline, reading
...
+
+NOTE: In cases when multiple threads are configured, `consumer_threads` is greater than one, a suffix is appended to
```
karenzone, line 366:
```suggestion
NOTE: In cases when multiple threads are configured, `consumer_threads` is greater than one, a suffix is appended to
```
```suggestion
NOTE: In cases when multiple threads are configured and `consumer_threads` is greater than one, a suffix is appended to
```

karenzone (commented):
I left some comments for your consideration.  In places, I made some guesses rather than asking questions to help us reduce the back-and-forth of multiple review cycles.  Please keep me honest, and let me know if you'd like to discuss.  :-)

docs/input-kafka.asciidoc
andsel: Yes you guessed right :-) I thinks that instead of
```
instance to be removed
```
would be more appropriate
```
instance to be disconnected
```

PR #137, opened 2023-02-02
Adds schema registry's truststore and keystore settings

### LINKED ISSUE #101: Confluent avro using https + user / password not working
**Logstash information**:

Please include the following information:

1. Logstash version: 7.13.3
2. Logstash installation source: Docker
3. How is Logstash being run: Docker
4. How was the Logstash Plugin installed:
Upgraded to latest stable using **logstash-plugin update logstash-integration-kafka** (v10.8.1)

**JVM** (e.g. `java -version`): Bundled JDK

**OS version** (`uname -a` if on a Unix-like system): 3.10.0-1160.24.1.el7.x86_64 #1 SMP Thu Mar 25 21:21:56 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

**Description of the problem including expected versus actual behavior**: 
Using HTTPS + auth does not work. Switching to HTTP with auth works however. 

**Steps to reproduce**:

```
input {
  kafka {
        bootstrap_servers => "mybroker.kafka:9094"
        group_id => "my_group"
        client_id => "my_client"
        topics => "mytopic"
        schema_registry_url => "https://my.registry:8081/"
        schema_registry_key => "myuser"
        schema_registry_secret => "mypassword"
        codec => "json"
        decorate_events => true
        auto_offset_reset => "earliest"
        jaas_path => "/path/to/kafka-jaas-input.conf"
        sasl_mechanism => "SCRAM-SHA-512"
        security_protocol => "SASL_SSL"
        ssl_truststore_location => "/path/to/certs/my.truststore.jks"
        ssl_truststore_password => "REDACTED"
        key_deserializer_class => "org.apache.kafka.common.serialization.ByteArrayDeserializer"
    }
}
```
**Provide logs (if relevant)**:
Using bundled version with 7.13.3 : 
>Pipeline error {:pipeline_id=>"main", :exception=>#<LogStash::ConfigurationError: Schema registry service doesn't respond, error: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target>

Using latest (10.8.1) version of this plugin. https + authentication: 
> [2021-07-15T13:17:49,790][ERROR][logstash.inputs.kafka    ] Unable to poll Kafka consumer {:kafka_error_message=>org.apache.kafka.common.errors.SerializationException: Error deserializing key/value for partition MYTOPIC-0 at offset 23815. If needed, please seek past the record to continue consumption., :cause=>org.apache.kafka.common.errors.SerializationException: Error retrieving Avro unknown schema for id 7}

From the errors I had with 7.11.1 I Highly suspect logstash does not pass the auth header (see this [SO article for more info](https://stackoverflow.com/questions/67586950/can-not-connect-logstash-to-confluent-schema-registry))

Using latest (10.8.1) version of this plugin, http (Haproxy for SSL termination)  + authentication **It works!**:
HaProxy Logs showing 200 http code **bolded**
>Jul 15 16:34:47 localhost haproxy[3142]: <redacted_IP>:40080 [15/Jul/2021:16:34:47.332] registry servers/srv1 0/0/65/24/89 **200** 5327 - - ---- 3/3/0/0/0 0/0 "GET /schemas/ids/7?fetchMaxId=false HTTP/1.1"

My Haproxy conf for this. 

```
frontend registry
    mode http
    bind :80
    default_backend servers


backend servers
    mode http
    server srv1 my.registry.intres:8081 ssl verify none
```

Can't test right now with an https frontend as generating a cert in my org takes a bit of time. Will update with results if I can.

andsel:
## Release notes
Add schema registry's setting for keystore and truststore.

## What does this PR do?

This commit mainly exposes `location` `password` and `type` settings for schema registry's secret and key stores.
It brings those configuration options, if available, and directly forward down to the Kafka's SerDes library and Manticore client.
Introduces a script named `setup_keystore_and_truststore.sh` to setup keystore and truststore used in integration tests.
Furthermore it reworks a little bit the bash scripts to setup Kafka and Schema Registry in integration test to avoid download of artifacts if they are already locally downloaded.

## Why is it important/What is the impact to the user?

With this PR the user is able to configure trust and key stores to be used to connect and interact specifically with Schema Registry.
These stores are different from the one configured for the Kafka client.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- ~~[ ] I have made corresponding change to the default configuration files (and/or docker env variables)~~
- [x] I have added tests that prove my fix is effective or that my feature works

## Author's Checklist

- [x] setup schema registry with separate keystore, import the certificate into client's truststore and verify that it can the plugin can connect to it running a pipeline.

## How to test this PR locally
To test this PR a couple of security settings has to be done. The best way to setup Kafka and Schema registry with key store is to launch the `kafka_test_setup.sh` and `./start_schema_registry.sh`. The script has the job to create keystore for schema registry, export it certificate and install on a truststore used by the client pipeline. All this security artifacts are available under the`./tls_repository` directory in plugin's clone.
1. clone the repo and launch:
```bash
./kafka_test_setup.sh
./start_schema_registry.sh
```
2. install a schema in the registry, for example
```
curl --location --request POST 'http://localhost:8081/subjects/logstash_test-value/versions' \
--header 'Content-Type: application/vnd.schemaregistry.v1+json' \
--data-raw '{"schema": "{\"namespace\": \"io.confluent.examples.clients.basicavro\", \"type\": \"record\", \"name\": \"Payment\", \"fields\": [     {\"name\": \"id\", \"type\": \"string\"},     {\"name\": \"amount\", \"type\": \"double\"} ]}"}'
```
  - if you want to use the HTTPS, add to the `curl`
```
--cacert /path/to/logstash-integration-kafka/tls/schema_reg_certificate.pem   
```
 -  point to port `8083` and switch protocol to`https`
3. in a local Logstash configure the Gemfile to point to the plugin:
  -  adds in `Gemfile`
```
gem "logstash-integration-kafka", :path => "/path/to/logstash-integration-kafka"
```
  - install it
```bash
bin/logstash-plugin install --no-verify
```
4. execute a pipeline with leverages the schema registry trust store setting:
```
input {
  kafka {
    bootstrap_servers => ["localhost:9092"]
    topics => ["logstash_test"]
    group_id => "logstash"
    consumer_threads => 2
    schema_registry_url => "https://localhost:8083/"
    schema_registry_ssl_truststore_location => "/path/to/logstash-integration-kafka/tls/clienttruststore.jks"
    schema_registry_ssl_truststore_password => "changeit"
    key_deserializer_class => "org.apache.kafka.common.serialization.ByteArrayDeserializer"
  }
}


output {
  stdout {
    codec => rubydebug
  }
}
``` 
5. verify no errors happen in log

## Related issues

- Closes #101 
- Relates https://github.com/confluentinc/schema-registry/pull/957

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -135,10 +135,10 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-schema_registry_key>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_proxy>> |<<uri,uri>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_secret>> |<<string,string>>|No
-| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_location>> |a valid filesystem path|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_path>> |a valid filesystem path|No
 | <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_password>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_type>> |<<string,string>>, one of `["jks", "PKCS12"]`|No
-| <<plugins-{type}s-{plugin}-schema_registry_ssl_truststore_location>> |a valid filesystem path|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_truststore_path>> |a valid filesystem path|No
 | <<plugins-{type}s-{plugin}-schema_registry_ssl_truststore_password>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_ssl_truststore_type>> |<<string,string>>, one of `["jks", "PKCS12"]`|No
 | <<plugins-{type}s-{plugin}-schema_registry_url>> |<<uri,uri>>|No
@@ -604,8 +604,8 @@ Set the address of a forward HTTP proxy. An empty string is treated as if proxy 
 
 Set the password for basic authorization to access remote Schema Registry.
 
-[id="plugins-{type}s-{plugin}-schema_registry_ssl_keystore_location"]
-===== `schema_registry_ssl_keystore_location`
+[id="plugins-{type}s-{plugin}-schema_registry_ssl_keystore_path"]
+===== `schema_registry_ssl_keystore_path`
 
 * Value type is <<path,path>>
 * There is no default value for this setting.
@@ -628,8 +628,8 @@ If schema registry authentication is required, this setting stores the keystore 
 
 The format of the keystore file. It must be either `jks` or `PKCS12`.
 
-[id="plugins-{type}s-{plugin}-schema_registry_ssl_truststore_location"]
-===== `schema_registry_ssl_truststore_location`
+[id="plugins-{type}s-{plugin}-schema_registry_ssl_truststore_path"]
+===== `schema_registry_ssl_truststore_path`
 
 * Value type is <<path,path>>
 * There is no default value for this setting.
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -135,6 +135,12 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-schema_registry_key>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_proxy>> |<<uri,uri>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_secret>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_location>> |a valid filesystem path|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_password>> |<<password,password>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_type>> |<<string,string>>, one of `["jks", "PKCS12"]`|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_truststore_location>> |a valid filesystem path|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_truststore_password>> |<<password,password>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_truststore_type>> |<<string,string>>, one of `["jks", "PKCS12"]`|No
 | <<plugins-{type}s-{plugin}-schema_registry_url>> |<<uri,uri>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_validation>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-security_protocol>> |<<string,string>>, one of `["PLAINTEXT", "SSL", "SASL_PLAINTEXT", "SASL_SSL"]`|No
@@ -597,6 +603,54 @@ Set the address of a forward HTTP proxy. An empty string is treated as if proxy 
 * There is no default value for this setting.
 
 Set the password for basic authorization to access remote Schema Registry.
+
+[id="plugins-{type}s-{plugin}-schema_registry_ssl_keystore_location"]
+===== `schema_registry_ssl_keystore_location`
+
+* Value type is <<path,path>>
+* There is no default value for this setting.
+
+If schema registry client authentication is required, this setting stores the keystore path.
+
+[id="plugins-{type}s-{plugin}-schema_registry_ssl_keystore_password"]
+===== `schema_registry_ssl_keystore_password`
+
+* Value type is <<password,password>>
+* There is no default value for this setting.
+
+If schema registry authentication is required, this setting stores the keystore password.
+
+[id="plugins-{type}s-{plugin}-schema_registry_ssl_keystore_type"]
+===== `schema_registry_ssl_keystore_type`
+
+* Value type is <<string,string>>
+* There is no default value for this setting.
+
+The format of the keystore file. It must be either `jks` or `PKCS12`.
+
+[id="plugins-{type}s-{plugin}-schema_registry_ssl_truststore_location"]
+===== `schema_registry_ssl_truststore_location`
+
+* Value type is <<path,path>>
+* There is no default value for this setting.
+
+The truststore path to validate the schema registry's certificate.
+
+[id="plugins-{type}s-{plugin}-schema_registry_ssl_truststore_password"]
+===== `schema_registry_ssl_truststore_password`
+
+* Value type is <<password,password>>
+* There is no default value for this setting.
+
+The schema registry truststore password.
+
+[id="plugins-{type}s-{plugin}-schema_registry_ssl_truststore_type"]
+===== `schema_registry_ssl_truststore_type`
+
+* Value type is <<string,string>>
+* There is no default value for this setting.
+
+The format of the schema registry's truststore file. It must be either `jks` or `PKCS12`.
 
 [id="plugins-{type}s-{plugin}-schema_registry_url"]
 ===== `schema_registry_url`
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -135,10 +135,10 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-schema_registry_key>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_proxy>> |<<uri,uri>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_secret>> |<<string,string>>|No
-| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_path>> |a valid filesystem path|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_location>> |a valid filesystem path|No
 | <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_password>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_type>> |<<string,string>>, one of `["jks", "PKCS12"]`|No
-| <<plugins-{type}s-{plugin}-schema_registry_ssl_truststore_path>> |a valid filesystem path|No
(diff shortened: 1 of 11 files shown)


lib/logstash/plugin_mixins/kafka/avro_schema_registry.rb
```diff
@@ -24,6 +24,24 @@ def setup_schema_registry_config
       # This option permits to define a proxy to be used to reach the schema registry service instance.
       config :schema_registry_proxy, :validate => :uri
 
+      # If schema registry client authentication is required, this setting stores the keystore path.
+      config :schema_registry_ssl_keystore_location, :validate => :string
```
edmocosta, line 28:
```suggestion
      config :schema_registry_ssl_keystore_path, :validate => :string
```

Do you think we should stick to the current approach, making config names as close as possible to Kafka's properties, or should we (kind of) apply the new plugin's naming convention for SSL? if we decide to keep it close to Kafka, IMO, we should also exclude this plugin from the upcoming standardization plans.

andsel: This naming was chosen to reflect the existing `ssl_keystore_location` but I agree, I'll adopt the new TLS naming convention

andsel: Suggestion applied

edmocosta: Actually, my comment's main intention was to discuss the SSL standardization for this plugin and if we should follow or not the new SSL naming convention here.

Considering this plugin's settings are already aligned with Kafka properties names, and that those settings are very particular and well-known by Kafka's users, I would keep your previous version, matching Kafka's names.
We've discussed the SSL standardization offline (@jsvd @yaauie @roaksoax)  and agreed on skipping (at least for now) Kafka plugins due to that reason.

If you agree, I think we could roll back your last commit and keep it as it's.

andsel: I think that adding new settings is better to avoid debt, and stick immediately with new naming conventions.
So if we all agree on this, we can keep this naming for new setting. A follow up PR will be created deprecate the existing ones and introduce the new naming here.

docs/input-kafka.asciidoc
```diff
@@ -135,6 +135,12 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-schema_registry_key>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_proxy>> |<<uri,uri>>|No
 | <<plugins-{type}s-{plugin}-schema_registry_secret>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_path>> |a valid filesystem path|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_password>> |<<password,password>>|No
+| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_type>> |<<string,string>>, one of `["jks", "PKCS12"]`|No
```
jsvd, line 140: are we ignore case for these settings? otherwise maybe we should stick to upper case on both?

```suggestion
| <<plugins-{type}s-{plugin}-schema_registry_ssl_keystore_type>> |<<string,string>>, one of `["JKS", "PKCS12"]`|No
```

kafka_test_setup.sh
```diff
@@ -24,9 +28,24 @@ echo "Starting Kafka broker"
 build/kafka/bin/kafka-server-start.sh -daemon build/kafka/config/server.properties --override advertised.host.name=127.0.0.1 --override log.dirs="${PWD}/build/kafka-logs"
 sleep 10
 
-echo "Downloading Confluent Platform"
-curl -s -o build/confluent_platform.tar.gz http://packages.confluent.io/archive/5.5/confluent-community-5.5.1-2.12.tar.gz
+echo "Setup Confluent Platform"
+if [ ! -e confluent-community-5.5.1-2.12.tar.gz ]; then
```
jsvd, line 32: Let's abstract this version of cp to the top of the file, like we do with KAFKA_VERSION

docs/input-kafka.asciidoc
andsel: Went for lowercase to adhere to the rest of the plugin: c70ef26b2f27f4c817efc234348d2723ade1c727

lib/logstash/plugin_mixins/kafka/avro_schema_registry.rb
```diff
@@ -24,6 +24,24 @@ def setup_schema_registry_config
...
       config :schema_registry_proxy, :validate => :uri
 
+      # If schema registry client authentication is required, this setting stores the keystore path.
+      config :schema_registry_ssl_keystore_path, :validate => :string
+
+      # The keystore password.
+      config :schema_registry_ssl_keystore_password, :validate => :password
+
+      # The keystore type
+      config :schema_registry_ssl_keystore_type, :validate => :string, :default => "jks"
```
mashhurs, line 34: Docs says, we allow either `jks` or `PKCS12`.

```suggestion
      config :schema_registry_ssl_keystore_type, :validate => ['jks', 'PKCS12'], :default => "jks"
```

lib/logstash/plugin_mixins/kafka/avro_schema_registry.rb
```diff
@@ -24,6 +24,24 @@ def setup_schema_registry_config
...
+
+      # The JKS truststore path to validate the Schema Registry's certificate.
+      config :schema_registry_ssl_truststore_path, :validate => :string
+
+      # The truststore password.
+      config :schema_registry_ssl_truststore_password, :validate => :password
+
+      # The truststore type
+      config :schema_registry_ssl_truststore_type, :validate => :string, :default => "jks"
```
mashhurs, line 43: Same as `schema_registry_ssl_keystore_type`:
```suggestion
      config :schema_registry_ssl_truststore_type, :validate => ['jks', 'PKCS12'], :default => "jks"
```

lib/logstash/plugin_mixins/kafka/avro_schema_registry.rb
```diff
@@ -68,6 +86,23 @@ def check_for_schema_registry_connectivity_and_subjects
       if schema_registry_key and !schema_registry_key.empty?
         options[:auth] = {:user => schema_registry_key, :password => schema_registry_secret.value}
       end
+      if schema_registry_ssl_truststore_path and !schema_registry_ssl_truststore_path.empty?
+        options[:ssl] = {} unless options.key?(:ssl)
+        options[:ssl].merge!({
+          :truststore => schema_registry_ssl_truststore_path,
+          :truststore_password => schema_registry_ssl_truststore_password.value,
```
mashhurs, line 93: `schema_registry_ssl_truststore_password` might be `nil` as it doesn't have a default value OR don't we need to fail to accept empty password?
```suggestion
          :truststore_password => schema_registry_ssl_truststore_password&.value,
```

lib/logstash/plugin_mixins/kafka/avro_schema_registry.rb
```diff
@@ -68,6 +86,23 @@ def check_for_schema_registry_connectivity_and_subjects
...
+          :truststore_type => schema_registry_ssl_truststore_type,
+        })
+      end
+      if schema_registry_ssl_keystore_path and !schema_registry_ssl_keystore_path.empty?
+        options[:ssl] = {} unless options.key? :ssl
+        options[:ssl].merge!({
+          :keystore => schema_registry_ssl_keystore_path,
+          :keystore_password => schema_registry_ssl_keystore_password.value,
```
mashhurs, line 101: Same as `schema_registry_ssl_truststore_password `
```suggestion
          :keystore_password => schema_registry_ssl_keystore_password&.value,
```

kafka_test_setup.sh
andsel: Done with commit ba0274779d24fd0d4a1935d7c8fb8dbb75e53e27

lib/logstash/plugin_mixins/kafka/avro_schema_registry.rb
andsel: Good point, maybe for all this settings is better to use the same pattern used at https://github.com/logstash-plugins/logstash-integration-kafka/blob/0462434fc6cf2c1de0d9314bbeb0c0840473d223/lib/logstash/plugin_mixins/kafka/common.rb#L23

andsel: solved by 4914eb4eb3ddde5094fde9d960ebc697be6fad34

lib/logstash/plugin_mixins/kafka/avro_schema_registry.rb
andsel: solved by 4914eb4eb3ddde5094fde9d960ebc697be6fad34

setup_keystore_and_truststore.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Setup Schema Registry keystore and Kafka's schema registry client's truststore
+set -ex
+
```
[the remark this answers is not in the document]
jsvd: My only concern is that if we do have more than 1 script, we have a reasonable confidence that the user is invoking the right script, and if script B can't be run before script A, we have some precondition check in script B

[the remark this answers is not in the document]
andsel: Specifically to this case, creation of security credentials, the script is invoked by the other `kafka_test_setup.sh`, and it's not intended to be invoked directly by the user. However if the user does, then the files it previously created are simply recreated.

jsvd (approved):
LGTM, everything seems to be working correctly, thanks!

lib/logstash/plugin_mixins/kafka/avro_schema_registry.rb
edmocosta: Did we get an agreement on this? @jsvd?

jsvd: Ah, I missed this one, thanks @edmocosta.
@andsel we decided we should leave the Kafka integration outside of the TLS normalization process since there are naming conventions users of Kafka will be familiar with.
So I'd prefer us to use `_location` instead of `_path`, sorry for the back and forth.

andsel: Reverted with 5dc98c5b586eea826336ac58ae9d8c9ad3dced25

mashhurs (approved):
💯

Issue #155, opened 2024-01-08
"Can't modify frozen String" For Kafka Tombstone


jpohl41:
**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`)
`8.11.3`
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
`Dockerized .tar.gz`
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
`supervisord`
4. How was the Logstash Plugin installed
`Default install`

**JVM** (e.g. `java -version`):
`openjdk 17.0.9`

**OS version** (`uname -a` if on a Unix-like system):
`Alpine linux docker`

**Description of the problem including expected versus actual behavior**:
**Actual Behavior:**
When sending a Kafka "tombstone" record (a.k.a. a record with a `null` value) I am seeing the following failure:
```
2023-11-21T00:01:36,761][ERROR][org.logstash.Logstash    ] uncaught exception (in thread kafka-input-worker-logstash-jobs-kafka-topic-0)
org.jruby.exceptions.FrozenError: (FrozenError) can't modify frozen String
        at org.jruby.RubyString.force_encoding(org/jruby/RubyString.java:6659) ~[jruby.jar:?]
        at org.jruby.RubyString.force_encoding(org/jruby/RubyString.java:6655) ~[jruby.jar:?]
        at opt.logstash.logstash_minus_core.lib.logstash.util.charset.convert(/opt/logstash/logstash-core/lib/logstash/util/charset.rb:29) ~[?:?]
        at opt.logstash.vendor.bundle.jruby.$3_dot_1_dot_0.gems.logstash_minus_codec_minus_plain_minus_3_dot_1_dot_0.lib.logstash.codecs.plain.decode(/opt/logstash/vendor/bundle/jruby/3.1.0/gems/logstash-codec-plain-3.1.0/lib/logstash/codecs/plain.rb:50) ~[?:?]
        at opt.logstash.logstash_minus_core.lib.logstash.codecs.delegator.decode(/opt/logstash/logstash-core/lib/logstash/codecs/delegator.rb:62) ~[?:?]
        at org.logstash.instrument.metrics.AbstractSimpleMetricExt.time(org/logstash/instrument/metrics/AbstractSimpleMetricExt.java:74) ~[logstash-core.jar:?]
[...]

PR #156, opened 2024-01-09
Fix "Can't modify frozen string" error when record value is `nil`

edmocosta:
This PR fixes the issue caused by the Ruby change https://bugs.ruby-lang.org/issues/16150, which made `boolean`, module name & `nil` `.to_s` values frozen by default, making this plugin unable to process Kafka "tombstone" records (a.k.a. a record with a `null` value).

#### How to test it locally?

1 - Set up the Kafka test environment:
```shell
./kafka_test_setup.sh
``` 
2 - Run the following pipeline config:
```yaml
 - pipeline.id: kafka
   config.string: |
     input {
       kafka {
         topics => "logstash_integration_topic_plain"
       }
     }

    output { stdout { codec => rubydebug { metadata => true } } }
```
3 - Produce a message with a null content:
```shell
echo "NULL" | build/kafka/bin/kafka-console-producer.sh --topic logstash_integration_topic_plain --broker-list localhost:9092 --property null.marker=NULL
```

Using the `main` branch code, it should terminate the pipeline and log an error similar to:

```
Exception in thread "kafka-input-worker-logstash-0" org.jruby.exceptions.FrozenError: (FrozenError) can't modify frozen String
	at org.jruby.RubyString.force_encoding(org/jruby/RubyString.java:6667)
	at org.jruby.RubyString.force_encoding(org/jruby/RubyString.java:6663)
```

The error should not happen using the fixed version.

---
Closes: https://github.com/logstash-plugins/logstash-integration-kafka/issues/155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
+## 11.3.3
+  - Fix "Can't modify frozen string" error when record value is `nil` (tombstones) [#155](https://github.com/logstash-plugins/logstash-integration-kafka/pull/155)
+
 ## 11.3.2
-  - Fix: update Avro library [#150](https://api.github.com/repos/logstash-plugins/logstash-integration-kafka/pulls/150)
+  - Fix: update Avro library [#150](https://github.com/logstash-plugins/logstash-integration-kafka/pull/150)
 
 ## 11.3.1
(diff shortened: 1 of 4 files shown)


jsvd (approved):
LGTM, thanks for the quick fix.

PR #162, opened 2024-03-22
Kafka Output support manipulating kafka message headers

LaithSiriani:
- Add the ability to set headers in the output kafka message
- Add unittest and integration test case to verify

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.4.0
+  - Add support for setting Kafka message headers in output plugin [#162](https://github.com/logstash-plugins/logstash-integration-kafka/pull/162)
+
 ## 11.3.4
   - Fix "retries" and "value_serializer" error handling in output plugin (#160) [#160](https://github.com/logstash-plugins/logstash-integration-kafka/pull/160)
 
--- a/docs/output-kafka.asciidoc
+++ b/docs/output-kafka.asciidoc
@@ -91,6 +91,7 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-key_serializer>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-linger_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-max_request_size>> |<<number,number>>|No
+| <<plugins-{type}s-{plugin}-message_header>> |<<hash,hash>>|No
 | <<plugins-{type}s-{plugin}-message_key>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-metadata_fetch_timeout_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-metadata_max_age_ms>> |<<number,number>>|No
@@ -274,6 +275,22 @@ to allow other records to be sent so that the sends can be batched together.
   * Default value is `1048576` (1MB).
 
 The maximum size of a request
+
+[id="plugins-{type}s-{plugin}-message_headers"]
+===== `message_headers`
+
+  * Value type is <<hash,hash>>
+  ** Keys are header names, and must be <<string,string>>
+  ** Values are header values, and must be <<string,string>>
+  ** Values support interpolation from event field values
+  * There is no default value for this setting.
+
+A map of key value pairs, each corresponding to a header name and its value respectively.
+Example:
+[source,ruby]
+----------------------------------
+    message_header => { "event_timestamp" => "%{@timestamp}" }
+----------------------------------
 
 [id="plugins-{type}s-{plugin}-message_key"]
 ===== `message_key` 
--- a/lib/logstash/outputs/kafka.rb
+++ b/lib/logstash/outputs/kafka.rb
@@ -106,6 +106,8 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
   config :max_request_size, :validate => :number, :default => 1_048_576 # (1MB) Kafka default
   # The key for the message
   config :message_key, :validate => :string
+  # Headers added to kafka message in the form of key-value pairs
+  config :message_headers, :validate => :hash, :default => {}
   # the timeout setting for initial metadata request to fetch topic metadata.
   config :metadata_fetch_timeout_ms, :validate => :number, :default => 60_000
   # Partitioner to use - can be `default`, `uniform_sticky`, `round_robin` or a fully qualified class name of a custom partitioner.
@@ -203,6 +205,11 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
       end
     else
       raise LogStash::ConfigurationError, "'value_serializer' only supports org.apache.kafka.common.serialization.ByteArraySerializer and org.apache.kafka.common.serialization.StringSerializer"
+    end
+    @message_headers.each do |key, value|
+      if !key.is_a? String
+        raise LogStash::ConfigurationError, "'message_headers' contains a key that is not a string!"
+      end
     end
     @producer = create_producer
   end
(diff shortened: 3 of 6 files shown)


LaithSiriani:
Due to previous failure in the Travis CI build, I modified the commit and uploaded a fix for some flakiness in the "when KafkaProducer#send() raises a non-retriable exception" test case

LaithSiriani:
@roaksoax Hi, would it be possible to review this pull request? What would be needed to include this feature?

LaithSiriani:
@mashhurs  Sorry for the possibly unnecessary involvement, I saw you had reviewed the last pull request. What is the way of working to get a code review for the pull request?

lib/logstash/outputs/kafka.rb
```diff
@@ -312,8 +314,18 @@ def close
...
     if @message_key.nil?
       record = ProducerRecord.new(event.sprintf(@topic_id), serialized_data)
+      if !@message_headers.nil?
+        @message_headers.each do |key, value|
+          record.headers().add(event.sprintf(key), event.sprintf(value).to_java_bytes)
+        end
+      end
     else
       record = ProducerRecord.new(event.sprintf(@topic_id), event.sprintf(@message_key), serialized_data)
+      if !@message_headers.nil?
```
mashhurs, line 324: - If you make `message_headers` config with a default value, you can avoid always `nil?` checks.
- It looks newly introduced lines are identical, so we can move the logic to out of this scope (just before `prepare(record)`)

logstash-integration-kafka.gemspec
```diff
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = 'logstash-integration-kafka'
-  s.version         = '11.3.4'
+  s.version         = '11.4.0'
```
mashhurs, line 3: Kudos for this 🎉

spec/integration/outputs/kafka_spec.rb
```diff
@@ -191,6 +191,25 @@
     end
   end
 
+  context 'when setting message_headers' do
+    let(:num_events) { 10 }
+    let(:test_topic) { 'logstash_integration_topic4' }
+
+    before :each do
+      config = base_config.merge({"topic_id" => test_topic, "message_headers" => {"k1" => "v1"}})
```
mashhurs, line 199: can you also add the case to validate the `event.sprintf`?

spec/unit/outputs/kafka_spec.rb
```diff
@@ -103,7 +113,7 @@
   end
 
   context "when KafkaProducer#send() raises a non-retriable exception" do
-    let(:failcount) { (rand * 10).to_i }
+    let(:failcount) { (rand * 10 + 1).to_i }
```
mashhurs, line 116: not sure if this is a relative change

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 11.4.0
+  - Add support for setting Kafka message headers in output plugin
```
mashhurs, line 2:
```suggestion
  - Add support for setting Kafka message headers in output plugin [#162](https://github.com/logstash-plugins/logstash-integration-kafka/pull/162)
```

CHANGELOG.md
```diff
@@ -23,7 +26,7 @@
  - Added config `group_instance_id` to use the Kafka's consumer static membership feature [#135](https://github.com/logstash-plugins/logstash-integration-kafka/pull/135)
 
 ## 11.0.0
-  - Changed Kafka client to 3.3.1, requires Logstash >= 8.3.0.
```
mashhurs, line 26: Can you also please revert this format changes? We have a plan to make them formatted with a standard formatter (currently in progress). Thanks.

lib/logstash/outputs/kafka.rb
```diff
@@ -106,6 +106,8 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
   config :max_request_size, :validate => :number, :default => 1_048_576 # (1MB) Kafka default
   # The key for the message
   config :message_key, :validate => :string
+  # Headers added to kafka message in the form of key-value pairs
+  config :message_headers, :validate => :hash
```
mashhurs, line 110: Having a default value would be beneficial to avoid always `nil?` checks.
This also requires doc changes:
- add to the [params table](https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/docs/input-kafka.asciidoc?plain=1#L97) with alphabetical order
- explain it in the main doc content (obeys the order of the param options table)

mashhurs (commented):
Thanks for the contribution @LaithSiriani!
I have placed some comments and I do believe they are relatively easy to make changes. At the same time, I am thinking (not sure so far) if we can abstract this in a decorative ways like we have in [kafka-input](https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/lib/logstash/inputs/kafka.rb#L249)

spec/unit/outputs/kafka_spec.rb
LaithSiriani: It is not related to the goal of the changeset but I got annoyed that the first patchset uploaded failed due to the instability of this test case, I can revert the change if needed.

spec/integration/outputs/kafka_spec.rb
LaithSiriani: Done

lib/logstash/outputs/kafka.rb
LaithSiriani: Done

CHANGELOG.md
LaithSiriani: Done

docs/output-kafka.asciidoc
```diff
@@ -275,6 +276,14 @@ to allow other records to be sent so that the sends can be batched together.
...
 The maximum size of a request
 
+[id="plugins-{type}s-{plugin}-message_header"]
+===== `message_key`
+
+  * Value type is <<string,string>>
+  * There is no default value for this setting.
+
+The key for the message.
+
```
yaauie, line 286: This will need to be updated to reflect the actual details of the new field. It's defined as a hash (key/value map), so likely something like:

```suggestion
===== `message_header`

  * Value type is <<hash,hash>>
  ** Keys are header names, and must be <<string,string>>
  ** Values are header values, and must be <<string,string>>
  ** Both Keys and Values support interpolation from event field values
  * There is no default value for this setting.

The headers to send with the message, such as:

[source]
----
    message_header => { "event_timestamp" => "%{@timestamp}" }
----
```

yaauie (commented):
Overall I am +1 to the feature.

I've added docs-related suggestions, and believe that we would benefit from calling this new option `message_headers` (plural) because we are providing a mapping that can be plural, which aligns with things like the http output.

We may also benefit from checking the value during `Kafka#register` to ensure that both keys and values are strings in order to avoid runtime surprises (I wish our built-in validators handled this, but they do not).

If we don't have an explicit use-case for making the _keys_ variable, then I would suggest avoiding the sprintf on the keys (and changing the docs to reflect that only the _values_ are interpolated).

spec/unit/outputs/kafka_spec.rb
LaithSiriani: this test case started failing again after I reverted the change, should I re-introduce it?

docs/output-kafka.asciidoc
LaithSiriani: Hi @yaauie , the documentation was updated according to your proposal, the key is now a static string. and a type check for the keys in message_headers had been added to the register function.

LaithSiriani:
Hi @yaauie, @mashhurs , the comments had been addressed.

LaithSiriani:
Hi @mashhurs , the comment was explicitly about the key, which has been addressed, the value does have a use case to be interpolated by the sprintf

mashhurs (approved):
LGTM!

mashhurs:
New v11.4.0 version released with this change.

PR #169, opened 2024-05-16
Add default client_id of logstash to kafka output

strawgate:
client_id defaults to `logstash` on the input, we should also default it to `logstash` on the output

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 11.3.4
+  - Add default client_id of logstash to kafka output [#169](https://github.com/logstash-plugins/logstash-integration-kafka/pull/169)
+
 ## 11.4.1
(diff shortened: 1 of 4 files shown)


jsvd (changes requested):
Please make the accompanying documentation change in the [output docs](https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/docs/output-kafka.asciidoc?plain=1#L191-L195), similar to what we have in the [input docs](https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/docs/input-kafka.asciidoc?plain=1#L227-L231)

strawgate:
Oh, good call! Should be in now

jsvd (approved):
LGTM

jsvd:
11.4.2 has been published as https://rubygems.org/gems/logstash-integration-kafka/versions/11.4.2-java, which will be included in the next Logstash patch/minor release.

PR #172, opened 2024-07-01
Ability to enable/disable Auto Topic Creation in Kafka input plugin.

e-parth-pathak:
- The Logstash Kafka input plugin should have the ability to enable/disable the auto creation of topics from the Consumer (logstash). 
- Users who have enabled auto topic creation in Kafka so that topics are automatically created when a event is produced should also be allowed to enable OR disable the same from Logstash's end.  
- The downside of this autocreation function is that topics can also be created by a consumers. This makes it difficult to remove topics that no longer produce events. There wasn't a property for this in the logstash's Kafka input plugin documentation.

Documentation ref: https://cwiki.apache.org/confluence/display/KAFKA/KIP-361%3A+Add+Consumer+Configuration+to+Disable+Auto+Topic+Creation

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -290,7 +290,7 @@ This will add a field named `kafka` to the logstash event containing the followi
 ===== `auto_create_topics` 
 
   * Value type is <<boolean,boolean>>
-  * Default value is `true`
+  * Default value is `true`
 
 Controls whether the topic is automatically created when subscribing to a non-existent topic.
 A topic will be auto-created only if this configuration is set to `true` and auto-topic creation is enabled on the broker using `auto.create.topics.enable`;
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -285,6 +285,18 @@ This will add a field named `kafka` to the logstash event containing the followi
 * `offset`: The offset from the partition this message is associated with
 * `key`: A ByteBuffer containing the message key
 
+
+[id="plugins-{type}s-{plugin}-auto_create_topics"]
+===== `auto_create_topics` 
+
+  * Value type is <<boolean,boolean>>
+  * Default value is `true`
+
+Controls whether the topic is automatically created when subscribing to a non-existent topic.
+A topic will be auto-created only if this configuration is set to `true` and auto-topic creation is enabled on the broker using `auto.create.topics.enable`;
+otherwise auto-topic creation is not permitted. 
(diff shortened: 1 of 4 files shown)


docs/input-kafka.asciidoc
```diff
@@ -1,4 +1,4 @@
-:integration: kafka
+:uto_cintegration: kafka
```
jsvd, line 1:
```suggestion
:integration: kafka
```

lib/logstash/inputs/kafka.rb
```diff
@@ -410,6 +416,7 @@ def create_consumer(client_id, group_instance_id)
 
       props.put(kafka::AUTO_COMMIT_INTERVAL_MS_CONFIG, auto_commit_interval_ms.to_s) unless auto_commit_interval_ms.nil?
       props.put(kafka::AUTO_OFFSET_RESET_CONFIG, auto_offset_reset) unless auto_offset_reset.nil?
+      props.put(kafka::AUTO_CREATE_TOPICS, auto_create_topics) unless auto_create_topics.nil?
```
jsvd, line 419: Given the setting is called [allow.auto.create.topics](https://kafka.apache.org/documentation/#consumerconfigs_allow.auto.create.topics):

```suggestion
      props.put(kafka::ALLOW_AUTO_CREATE_TOPICS, auto_create_topics) unless auto_create_topics.nil?
```

jsvd (changes requested):

docs/input-kafka.asciidoc
```diff
@@ -285,6 +285,17 @@ This will add a field named `kafka` to the logstash event containing the followi
 * `offset`: The offset from the partition this message is associated with
 * `key`: A ByteBuffer containing the message key
 
+
+[id="plugins-{type}s-{plugin}-auto_create_topics"]
```
jsvd, line 289: Can you add the entry for this option on the list index at the top of the document on line 101?

```
[cols="<,<,<",options="header",]
|=======================================================================
|Setting |Input type|Required
| <<plugins-{type}s-{plugin}-auto_commit_interval_ms>> |<<number,number>>|No
>>> **insert here**
| <<plugins-{type}s-{plugin}-auto_offset_reset>> |<<string,string>>|No
| <<plugins-{type}s-{plugin}-bootstrap_servers>> |<<string,string>>|No
| <<plugins-{type}s-{plugin}-check_crcs>> |<<boolean,boolean>>|No
```

jsvd (approved):
LGTM

PR #126, opened 2022-09-07
Add support for AWS IAM authentication

korsosa:
AWS MSK has an optional IAM auth feature that enables to use AWS IAM policies for managing authentication/authorization to Kafka resources instead of Kafka ACLs. More information in [the AWS MSK documentation](https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html). 

This change
- includes the [AWS library for MSK IAM auth](https://github.com/aws/aws-msk-iam-auth) that provides a new SASL mechanism `AWS_MSK_IAM`
- adds a config option `sasl.client.callback.handler.class` to allow providing custom SASL callback handlers
- in previous plugin versions the transitive Jar dependencies were not exported. The AWS MSK IAM auth library is a jar without dependencies, this caused issues with finding various AWS classes. Instead of listing all the transitive AWS dependencies, this change also modifies `build.gradle` to export all Jar dependencies and add them to the jars file. If a transitive dependency should not be exported, the `transitive=false` flag can be used.

Tested both input and output on an AWS EC2 machine with AWS MSK (Kafka version 2.6.2) and Logstash 8.2.3.

Usage as input/output:
```
 kafka {
    topic_id => "yourtopic"
    bootstrap_servers => "yourbroker"
    security_protocol => "SASL_SSL"
    sasl_mechanism => "AWS_MSK_IAM"
    sasl_jaas_config => "software.amazon.msk.auth.iam.IAMLoginModule required;"
    sasl_client_callback_handler_class => "software.amazon.msk.auth.iam.IAMClientCallbackHandler"
  }
```

This PR also fixes the issue https://github.com/logstash-plugins/logstash-integration-kafka/issues/124

Closes: #124

PR #177, opened 2024-08-19
Exposes `sasl.client.callback.handler.class` Kafka client setting to select a specific SASL Handler class

andsel:
This is the first part of PR splitting of #126.
It exposes the SASL client callback class setting to the Logstash configuration.


mashhurs (approved):
lgtm!

Issue #180, opened 2024-09-12
Support oauthbearer 


hgword:
Some oauth config is missing,
https://docs.confluent.io/cloud/current/security/authenticate/workload-identities/identity-providers/oauth/configure-clients-oauth.html

![image](https://github.com/user-attachments/assets/22c3f069-c1db-4184-ba99-6455982a1982)

PR #181, opened 2024-09-12
Add missing oauth config (#180)

hgword:


donoghuc:
Thanks @hgword for your commits! I have reviewed and cherry-picked them into a branch and added some unit tests in https://github.com/logstash-plugins/logstash-integration-kafka/pull/189 Please have a look there and we will get a release out with these options.

PR #183, opened 2024-09-12
sasl.oauthbearer.token.endpoint.url Kafka client setting to set oauth…

roon-replica:
…bearer token endpoint url

like PR https://github.com/logstash-plugins/logstash-integration-kafka/pull/177, i also want to add sasl_oauthbearer_token_endpoint_url.


donoghuc:
Looks like this got folded in to https://github.com/logstash-plugins/logstash-integration-kafka/pull/183/files which will subsequently be included in https://github.com/logstash-plugins/logstash-integration-kafka/pull/189 . Thanks for the contribution :)

PR #189, opened 2025-01-06
Add fine tuned oath/sasl configuration to kafka client

donoghuc:
Add plugin configuration options for both input/output to support oauth and sasl authentication. 

This PR aggregates https://github.com/logstash-plugins/logstash-integration-kafka/pull/183 and https://github.com/logstash-plugins/logstash-integration-kafka/pull/181 and adds some unit tests. 

I found it difficult to spin up a valid test environment to fully test the integrations, but looking at the options in https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/config/SaslConfigs.java I think the added unit tests show that we are piping through the requested configuration to the kafka clients. 

Closes https://github.com/logstash-plugins/logstash-integration-kafka/issues/180

### DIFF
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -600,6 +600,7 @@ The SASL login callback handler class the specified SASL mechanism should use.
 [id="plugins-{type}s-{plugin}-sasl_login_read_timeout_ms"]
 ===== `sasl_login_read_timeout_ms`
   * Value type is <<number,number>>
+  * There is no default value for this setting.
 
 (optional) The duration, in milliseconds, for HTTPS read timeout.
 
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -131,6 +131,12 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-request_timeout_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-retry_backoff_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-sasl_client_callback_handler_class>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-sasl_oauthbearer_token_endpoint_url>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-sasl_oauthbearer_scope_claim_name>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_connect_timeout_ms>> |<<number,number>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_read_timeout_ms>> |<<number,number>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_retry_backoff_ms>> |<<number,number>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_retry_backoff_max_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-sasl_jaas_config>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-sasl_kerberos_service_name>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-sasl_mechanism>> |<<string,string>>|No
@@ -556,12 +562,54 @@ retries are exhausted.
 The amount of time to wait before attempting to retry a failed fetch request
 to a given topic partition. This avoids repeated fetching-and-failing in a tight loop.
 
-[id="plugins-{type}s-{plugin}-sasl_client_callback_handler_class""]
+[id="plugins-{type}s-{plugin}-sasl_client_callback_handler_class"]
 ===== `sasl_client_callback_handler_class`
-* Value type is <<string,string>>
-* There is no default value for this setting.
+  * Value type is <<string,string>>
+  * There is no default value for this setting.
 
 The SASL client callback handler class the specified SASL mechanism should use.
+
+[id="plugins-{type}s-{plugin}-sasl_oauthbearer_token_endpoint_url"]
+===== `sasl_oauthbearer_token_endpoint_url`
+  * Value type is <<string,string>>
+  * There is no default value for this setting.
+
+The URL for the OAuth 2.0 issuer token endpoint.
+
+[id="plugins-{type}s-{plugin}-sasl_oauthbearer_scope_claim_name"]
+===== `sasl_oauthbearer_scope_claim_name`
+  * Value type is <<string,string>>
+  * Default value is `"scope"`
+
+(optional) The override name of the scope claim.
+
+[id="plugins-{type}s-{plugin}-sasl_login_connect_timeout_ms"]
+===== `sasl_login_connect_timeout_ms`
+  * Value type is <<number,number>>
+  * Default value is `10000` milliseconds.
+
+(optional) The duration, in milliseconds, for HTTPS connect timeout
+
+[id="plugins-{type}s-{plugin}-sasl_login_read_timeout_ms"]
+===== `sasl_login_read_timeout_ms`
+  * Value type is <<number,number>>
+  * Default value is `10000` milliseconds.
+
+(optional) The duration, in milliseconds, for HTTPS read timeout.
+
+[id="plugins-{type}s-{plugin}-sasl_login_retry_backoff_ms"]
+===== `sasl_login_retry_backoff_ms`
+  * Value type is <<number,number>>
+  * Default value is `100` milliseconds.
+
+(optional) The duration, in milliseconds, to wait between HTTPS call attempts.
+
+[id="plugins-{type}s-{plugin}-sasl_login_retry_backoff_max_ms"]
+===== `sasl_login_retry_backoff_max_ms`
+  * Value type is <<number,number>>
+  * Default value is `10000` milliseconds.
+
+(optional) The maximum duration, in milliseconds, for HTTPS call attempts.
 
 [id="plugins-{type}s-{plugin}-sasl_jaas_config"]
 ===== `sasl_jaas_config` 
--- a/docs/input-kafka.asciidoc
+++ b/docs/input-kafka.asciidoc
@@ -594,14 +594,12 @@ The SASL login callback handler class the specified SASL mechanism should use.
 [id="plugins-{type}s-{plugin}-sasl_login_connect_timeout_ms"]
 ===== `sasl_login_connect_timeout_ms`
   * Value type is <<number,number>>
-  * Default value is `10000` milliseconds.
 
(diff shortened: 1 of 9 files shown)


lib/logstash/inputs/kafka.rb
```diff
@@ -210,6 +210,20 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
...
+  # SASL login callback handler class
+  config :sasl_login_callback_handler_class, :validate => :string
+  # (optional) The duration, in milliseconds, for HTTPS connect timeout
+  config :sasl_login_connect_timeout_ms, :validate => :number, :default => 10000
+  # (optional) The duration, in milliseconds, for HTTPS read timeout.
+  config :sasl_login_read_timeout_ms, :validate => :number, :default => 10000
+  # (optional) The duration, in milliseconds, to wait between HTTPS call attempts.
+  config :sasl_login_retry_backoff_ms, :validate => :number, :default => 100
+  # (optional) The maximum duration, in milliseconds, for HTTPS call attempts.
+  config :sasl_login_retry_backoff_max_ms, :validate => :number, :default => 10000
```
robbavey, line 226: Will these settings. have any effect for producers and/or consumers? I can see these defined in [broker configs](https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#sasl-login-refresh-buffer-seconds), but not for [producers](https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#sasl-login-callback-handler-class) or [consumers](https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#sasl-login-callback-handler-class)?

donoghuc: I think so, for example: https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#sasl-login-retry-backoff-max-ms Maybe i'm missing some nuance though in what you are trying to point out?

lib/logstash/plugin_mixins/kafka/common.rb
```diff
@@ -42,6 +42,13 @@ def set_sasl_config(props)
       props.put("sasl.kerberos.service.name", sasl_kerberos_service_name) unless sasl_kerberos_service_name.nil?
       props.put("sasl.jaas.config", sasl_jaas_config) unless sasl_jaas_config.nil?
       props.put("sasl.client.callback.handler.class", sasl_client_callback_handler_class) unless sasl_client_callback_handler_class.nil?
+      props.put("sasl.oauthbearer.token.endpoint.url", sasl_oauthbearer_token_endpoint_url) unless sasl_oauthbearer_token_endpoint_url.nil?
+      props.put("sasl.oauthbearer.scope.claim.name", sasl_oauthbearer_scope_claim_name) unless sasl_oauthbearer_scope_claim_name.nil?
+      props.put("sasl.login.callback.handler.class", sasl_login_callback_handler_class) unless sasl_login_callback_handler_class.nil?
+      props.put("sasl.login.connect.timeout.ms", sasl_login_connect_timeout_ms.to_s) unless sasl_login_connect_timeout_ms.nil?
```
mashhurs, line 48: It seems currently, `sasl.login.connect.timeout.ms`,  `sasl.login.read.timeout.ms` and `sasl.login.retry.backoff.*` apply only to OAUTHBEARER ([reference](https://kafka.apache.org/31/generated/admin_client_config.html#adminclientconfigs_sasl.login.retry.backoff.max.ms)). WDYT if we set them _only_ when `sasl.oauthbearer.*` params set?

lib/logstash/outputs/kafka.rb
```diff
@@ -149,6 +149,20 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
...
   # SASL client callback handler class
   config :sasl_client_callback_handler_class, :validate => :string
+  # The URL for the OAuth 2.0 issuer token endpoint.
+  config :sasl_oauthbearer_token_endpoint_url, :validate => :string
+  # (optional) The override name of the scope claim.
+  config :sasl_oauthbearer_scope_claim_name, :validate => :string, :default => 'scope'
+  # SASL login callback handler class
+  config :sasl_login_callback_handler_class, :validate => :string
+  # (optional) The duration, in milliseconds, for HTTPS connect timeout
+  config :sasl_login_connect_timeout_ms, :validate => :number, :default => 10000
```
mashhurs, line 159: Don't we need to align the default value to kafka client [`sasl.login.connect.timeout.ms`](https://kafka.apache.org/31/generated/admin_client_config.html#adminclientconfigs_sasl.login.connect.timeout.ms) param? Same question for the `sasl.login.read.timeout.ms` and `sasl.login.retry.*` params.

lib/logstash/plugin_mixins/kafka/common.rb
mashhurs: The references:
- https://kafka.apache.org/31/generated/producer_config.html
- https://kafka.apache.org/31/generated/consumer_config.html

lib/logstash/inputs/kafka.rb
robbavey: No, you're right. I'm not quite sure how I missed it. 🤦

lib/logstash/outputs/kafka.rb
donoghuc: Seams reasonable. Appears to follow the pattern elsewhere in this file. Aligned the new options with defaults (or lack of defauts).

lib/logstash/plugin_mixins/kafka/common.rb
donoghuc: I couldn't find any evidence of these values being set actually harming anything. Given the language in the docs 
> The (optional) value in milliseconds for the external authentication provider connection timeout. Currently applies only to OAUTHBEARER.

the use of the work "currently" makes me think that they may extend it further at a later time. Trying to get in the middle here and add conditional logic does not seem worth the risk of *not* just letting the underlying client lib handle the values.

docs/output-kafka.asciidoc
```diff
@@ -392,13 +399,60 @@ In versions prior to 10.5.0, any exception is retried indefinitely unless the `r
...
+[id="plugins-{type}s-{plugin}-sasl_login_callback_handler_class"]
+===== `sasl_login_callback_handler_class`
+  * Value type is <<string,string>>
+  * There is no default value for this setting.
+
+The SASL login callback handler class the specified SASL mechanism should use.
+
+[id="plugins-{type}s-{plugin}-sasl_login_connect_timeout_ms"]
+===== `sasl_login_connect_timeout_ms`
+  * Value type is <<number,number>>
```
mashhurs, line 432:
```suggestion
  * Value type is <<number,number>>
  * There is no default value for this setting.
```

docs/input-kafka.asciidoc
```diff
@@ -556,13 +563,60 @@ retries are exhausted.
...
+
+[id="plugins-{type}s-{plugin}-sasl_login_connect_timeout_ms"]
+===== `sasl_login_connect_timeout_ms`
+  * Value type is <<number,number>>
+
+(optional) The duration, in milliseconds, for HTTPS connect timeout
+
+[id="plugins-{type}s-{plugin}-sasl_login_read_timeout_ms"]
+===== `sasl_login_read_timeout_ms`
+  * Value type is <<number,number>>
```
mashhurs, line 602:
```suggestion
  * Value type is <<number,number>>
  * There is no default value for this setting.
```

docs/input-kafka.asciidoc
```diff
@@ -556,13 +563,60 @@ retries are exhausted.
...
+[id="plugins-{type}s-{plugin}-sasl_login_callback_handler_class"]
+===== `sasl_login_callback_handler_class`
+  * Value type is <<string,string>>
+  * There is no default value for this setting.
+
+The SASL login callback handler class the specified SASL mechanism should use.
+
+[id="plugins-{type}s-{plugin}-sasl_login_connect_timeout_ms"]
+===== `sasl_login_connect_timeout_ms`
+  * Value type is <<number,number>>
```
mashhurs, line 596: not necessarily but it would be nice to add no default value indication as we are adding to all params.
```suggestion
  * Value type is <<number,number>>
  * There is no default value for this setting.
```

docs/output-kafka.asciidoc
```diff
@@ -392,13 +399,60 @@ In versions prior to 10.5.0, any exception is retried indefinitely unless the `r
...
+
+(optional) The duration, in milliseconds, for HTTPS connect timeout
+
+[id="plugins-{type}s-{plugin}-sasl_login_read_timeout_ms"]
+===== `sasl_login_read_timeout_ms`
+  * Value type is <<number,number>>
```
mashhurs, line 438:
```suggestion
  * Value type is <<number,number>>
  * There is no default value for this setting.
```

lib/logstash/plugin_mixins/kafka/common.rb
mashhurs: Thanks for clarifying~!

lib/logstash/outputs/kafka.rb
mashhurs: Awesome, thank you so much!

mashhurs (approved):
LGTM!
Thank you so much Cas.
The last failed CI seems to me transient.

docs/input-kafka.asciidoc
```diff
@@ -131,6 +131,13 @@ See the https://kafka.apache.org/{kafka_client_doc}/documentation for more detai
 | <<plugins-{type}s-{plugin}-request_timeout_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-retry_backoff_ms>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-sasl_client_callback_handler_class>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-sasl_oauthbearer_token_endpoint_url>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-sasl_oauthbearer_scope_claim_name>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_callback_handler_class>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_connect_timeout_ms>> |<<number,number>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_read_timeout_ms>> |<<number,number>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_retry_backoff_ms>> |<<number,number>>|No
+| <<plugins-{type}s-{plugin}-sasl_login_retry_backoff_max_ms>> |<<number,number>>|No
```
karenzone, line 140: The one got by me. 
For next time:   Options should be in alpha order in both option list and descriptions.

karenzone (commented):
Left suggestion inline to use order options alphabetically.

Issue #195, opened 2025-06-13
Upgrade Kafka Plugin to Use Kafka Client 4.0


aydasraf:
## Summary

Upgrade  to Kafka Client **4.0**, enabling **[KIP‑848 new consumer rebalance protocol](https://www.confluent.io/blog/kip-848-consumer-rebalance-protocol/)**. 

This change removes “stop-the-world” pauses and supports true dynamic scaling of consumer groups—critical for resilient, low-latency ingestion pipelines.

## Context & Problem

Our pipelines rely heavily on Logstash + Kafka. With the 3.x client, any consumer restart or failure triggers a full-group rebalance, pausing processing for all consumers. Despite tuning and cooperative protocols, outages persist and dynamic scaling is effectively impossible without interruption.

**Kafka 4.0** introduces **KIP‑848**, a broker-driven, incremental rebalance protocol that:
- Eliminates global sync barriers—pausing only the partitions that are reassigned  
- Enables incremental adds/removals without disrupting other consumers  
- Simplifies client logic by migrating coordination to brokers

**Benefits**:
- No full-group stalls during consumer restarts
- Seamless, true elastic scaling
- Reduced operational burden and better SLA alignment
- Alignment with Kafka ecosystem (KRaft mode, ZooKeeper removal, modern Java usage)

PR #206, opened 2025-10-09
[11.x] Deprecate partitioner default

kaisecheng:
- Deprecated partitioner `default` and `uniform_sticky` options, as `DefaultPartitioner` and `UniformStickyPartitioner` have been removed in kafka-client 4.0.
- Updated Kafka version used in integration tests for faster CDN downloads.

Relates: https://github.com/logstash-plugins/logstash-integration-kafka/issues/195

### DIFF
--- a/lib/logstash/outputs/kafka.rb
+++ b/lib/logstash/outputs/kafka.rb
@@ -423,7 +423,7 @@ class LogStash::Outputs::Kafka < LogStash::Outputs::Base
   end
 
   def log_partitioner_warning(partitioner, class_name)
-    logger.warn("Producer `partitioner` is configured with the deprecated option `#{partitioner}`. " \
+    deprecation_logger.deprecated("Producer `partitioner` is configured with the deprecated option `#{partitioner}`. " \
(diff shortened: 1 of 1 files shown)


docs/output-kafka.asciidoc
```diff
@@ -349,14 +349,15 @@ The max time in milliseconds before a metadata refresh is forced.
...
 
-The default behavior is to hash the `message_key` of an event to get the partition.
-When no message key is present, the plugin picks a partition in a round-robin fashion.
+By not setting this value, the plugin uses the default strategy provided by the Kafka client.
 
-Available options for choosing a partitioning strategy are as follows:
+Available options are as follows:
 
-* `default` use the default partitioner as described above
+* `default` hashes the `message_key` of an event to get the partition. When no message key is present, the plugin picks a partition in a round-robin fashion.
```
jsvd:
```suggestion
* `default` hashes the `message_key` of an event to get the partition. When no message key is present, the plugin picks a partition in a round-robin fashion. Please note that this is a different strategy than the one used when `partitioner` is left unset.
```

PR #252, opened 2026-04-18
Add share group consumer

### LINKED ISSUE #251: Add share group consumers in kafka
Kafka 4.2 now has share groups to allow for virtual queues:

https://docs.confluent.io/platform/current/config-manage/kafka-queues.html

I would like this implemented in logstash, so that i can scale up or down logstash without worrying about topic partitions.

orangecola:
Fixes #251 

E2E Testing:
```
$ docker exec ls-test sh -c 'pkill -f logstash; rm -f /tmp/share_output.jsonl /tmp/ls.log; sleep 2' 2>&1
docker exec kafka-sg /opt/kafka/bin/kafka-topics.sh --bootstrap-server localhost:9092 --create --topic final-test --partitions 3 --replication-factor 1 2>&1
docker exec kafka-sg /opt/kafka/bin/kafka-configs.sh --bootstrap-server localhost:9092 --alter --add-config share.auto.offset.reset=earliest --entity-type groups --entity-name final-group 2>&1
for i in $(seq 1 50); do echo "final-$i"; done | docker exec -i kafka-sg /opt/kafka/bin/kafka-console-producer.sh --bootstrap-server localhost:9092 --topic final-test 2>&1 | tail -1

cat > /tmp/final.conf <<'EOF'
input {
  kafka {
    bootstrap_servers => "kafka-sg:9092"
    topics => ["final-test"]
    group_id => "final-group"
    consumer_mode => "share_group"
    consumer_threads => 3
    client_id => "final-smoke"
    codec => "plain"
    poll_timeout_ms => 500
  }
}
output {
  file { path => "/tmp/share_output.jsonl" codec => line { format => "%{message}" } }
}
EOF
docker cp /tmp/final.conf ls-test:/tmp/final.conf
docker exec -d ls-test sh -c '/usr/share/logstash/bin/logstash -f /tmp/final.conf --path.data /tmp/ls-data-final > /tmp/ls.log 2>&1'
for i in 1 2 3 4 5 6 7 8; do
  sleep 5
  count=$(docker exec ls-test sh -c 'test -f /tmp/share_output.jsonl && wc -l < /tmp/share_output.jsonl || echo 0' 2>&1 | tr -d ' ')
  echo "t=${i}x5s: received=$count"
  if [ "$count" = "50" ]; then break; fi
done
echo "=== SUMMARY ==="
docker exec ls-test sh -c 'echo "total=$(wc -l < /tmp/share_output.jsonl)  unique=$(sort -u /tmp/share_output.jsonl | wc -l)  duplicates=$(sort /tmp/share_output.jsonl | uniq -d | wc -l)"' 2>&1
echo "=== error check ==="
docker exec ls-test sh -c 'grep -cE "\[ERROR\]" /tmp/ls.log' 2>&1
docker exec ls-test sh -c 'grep -E "\[ERROR\]" /tmp/ls.log | grep -v -E "licensereader|monitoring"' 2>&1 | head -5

Created topic final-test.
Completed updating config for group final-group.
t=1x5s: received=37
t=2x5s: received=37
t=3x5s: received=37
t=4x5s: received=87
t=5x5s: received=87
t=6x5s: received=87
t=7x5s: received=87
t=8x5s: received=87
=== SUMMARY ===
total=87  unique=85  duplicates=1
=== error check ===
3
```


lib/logstash/inputs/kafka.rb
```diff
@@ -565,4 +616,133 @@ def header_with_value(header)
...
+          end
+        end
+      ensure
+        consumer.close
+      end
+    end
+  end
+
+  public
+  def do_poll_share(consumer)
```
alexcams: This function is almost the same as `do_poll`, but it avoids catching `FencedInstanceIdException`. Do you think it could be possible to use `do_poll` instead declare a new function? I think catching an exception that never fires is harmless.

lib/logstash/inputs/kafka.rb
```diff
@@ -565,4 +616,133 @@ def header_with_value(header)
...
+            commit_share_acknowledgements(consumer)
+          end
+        end
+      ensure
+        consumer.close
+      end
+    end
+  end
+
+  public
```
alexcams: It is only called by internal function, can it be on private scope?

docs/input-kafka.asciidoc
```diff
@@ -285,8 +321,40 @@ https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+
...
+partition assignment, so any number of consumers can scale independently without being
+limited by the partition count.
+
+To use share groups, your Kafka brokers must be version 4.0 or later and must have
+`group.share.enable=true` set in the broker configuration.
+
+[NOTE]
+====
+The following options are not applicable in `share_group` mode and will be ignored or
+raise a configuration error if set to non-default values:
```
alexcams: The only options that will raise an error in such case will be `group_protocol`, `group_instance_id` and `partition_assignment_strategy`. I think it is worth clarify it here.

lib/logstash/inputs/kafka.rb
```diff
@@ -565,4 +616,133 @@ def header_with_value(header)
...
+        "The following options are not compatible with share_group consumer_mode and must be removed: #{overridden.join(', ')}"
+    end
+
+    if schema_registry_url
+      raise LogStash::ConfigurationError,
+        "schema_registry_url is not supported with share_group consumer_mode"
+    end
+  end
+
+  def create_share_consumer(client_id)
```
alexcams, line 669: This function is very similar to `create_consumer`. Create a common function like `build_common_props` will reduce duplicated code. WDYT? :)

alexcams (changes requested):
Thanks for this contribution @orangecola! The overall implementation looks nice, just left some nits worth to check.
Could you also update [CHANGELOG.md](https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/CHANGELOG.md) file and bump version in [gempspec](https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/logstash-integration-kafka.gemspec#L3)?

lib/logstash/inputs/kafka.rb
orangecola: Removed do_poll_share; share group runner now calls the existing do_poll (the FencedInstanceIdException rescue is harmless when it never fires).

lib/logstash/inputs/kafka.rb
orangecola: Moot now that do_poll_share is gone — do_poll was already public.

docs/input-kafka.asciidoc
orangecola: Split the asciidoc note (and inline doc) into two paragraphs — the 3 options that raise vs. the rest that are silently ignored.

lib/logstash/inputs/kafka.rb
orangecola: Extracted build_common_props(client_id) shared by create_consumer and create_share_consumer.

orangecola:
Addressed comments and updated version files.

> Thanks for this contribution @orangecola! The overall implementation looks nice, just left some nits worth to check. Could you also update [CHANGELOG.md](https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/CHANGELOG.md) file and bump version in [gempspec](https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/logstash-integration-kafka.gemspec#L3)?

alexcams (commented):
Thanks for addressing the comments @orangecola! I've been testing this feature manually and I can confirm the share group consumer works correctly — records are distributed across threads without specific thread-partition assignment.

While checking the docs, I've seen a recommendation for `consumer_threads` config that states: 
```
Ideally you should have as many threads as the number of partitions for a perfect balance -- more threads than partitions means that some threads will be idle
```
in 3 places:
- http://github.com/logstash-plugins/logstash-integration-kafka/blob/main/docs/input-kafka.asciidoc?plain=1#L61
- https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/docs/input-kafka.asciidoc?plain=1#L294
- https://github.com/logstash-plugins/logstash-integration-kafka/blob/main/lib/logstash/inputs/kafka.rb#L103


Would you mind updating this to be consistent with the new share_group consumer mode? 
Thanks in advance!

alexcams (changes requested):
Requested changes in [last review](https://github.com/logstash-plugins/logstash-integration-kafka/pull/252#pullrequestreview-4181568178)

orangecola:
mb, didn't push changes.

alexcams (approved):
Thanks @orangecola!

lib/logstash/inputs/kafka.rb
```diff
@@ -565,4 +631,81 @@ def header_with_value(header)
     !header.nil? && !header.value.nil? && !header.key.nil?
   end
 
+  def validate_share_group_config!
```
mashhurs, line 653:
```
not_allowed_configs_with_share_group =  original_params.keys & %w( group_protocol group_instance_id partition_assignment_strategy enable_auto_commit auto_commit_interval_ms)

if not_allowed_configs_with_share_group.size > 1
  error_message = "#{not_allowed_configs_with_share_group} options cannot be used together when `consumer_mode => 'share_group'`"
  raise LogStash::ConfigurationError, error_message
end
```

lib/logstash/inputs/kafka.rb
```diff
@@ -281,6 +285,35 @@ class LogStash::Inputs::Kafka < LogStash::Inputs::Base
...
+  #   record is delivered to exactly one consumer. Consumers must acknowledge each record.
+  #   This mode uses `KafkaShareConsumer` internally and requires Kafka brokers 4.0 or later
+  #   with `group.share.enable=true` on the broker.
+  #
+  # In `share_group` mode the following options raise a configuration error if set to
+  # a non-default value: `group_protocol`, `group_instance_id`, `partition_assignment_strategy`.
+  # Schema Registry (`schema_registry_url`) is also not supported.
+  # The following options are silently ignored in `share_group` mode because they do not
+  # apply to the Share Group protocol: `heartbeat_interval_ms`, `session_timeout_ms`,
+  # `isolation_level`, `enable_auto_commit`, `auto_commit_interval_ms`, `auto_offset_reset`,
```
mashhurs, line 303: I will also pull, run on my local and review in detail but I would be really specific and not allow `enable_auto_commit`, `heartbeat_interval_ms`, etc... params if pipeline config contains. The `original_params.keys` provides the pipeline config param keys (I will comment below how it can be used).

mashhurs (changes requested):
I believe this feature will be 🚀. Thank you alot for the contribution.
The implementation details, docs and format look good to me as well (will review docs section with second round).

One critical concern is that when pipeline config contains the params which cannot be used with `share_group`, I would support raising a config error instead of silently ignoring. For example, the following config looks invalid to me:
```
input {
  kafka {
    enable_auto_commit => true
    consumer_mode => "share_group"
  }
}
```

orangecola:
Combined `schema_registry_url` with other field checks.

lib/logstash/inputs/kafka.rb
```diff
@@ -565,4 +642,72 @@ def header_with_value(header)
...
+      begin
+        codec_instance = @codec.clone
+        ack_accept = org.apache.kafka.clients.consumer.AcknowledgeType::ACCEPT
+        until stop?
+          records = do_poll(consumer)
+          unless records.empty?
+            records.each do |record|
+              begin
+                handle_record(record, codec_instance, logstash_queue)
+                consumer.acknowledge(record, ack_accept)
```
yaauie: If the plugin is being stopped, should we still try to handle all of the events from the last poll, or should we release them to be processed by another consumer?

Because pushing into the pipeline's queue is effectively a blocking operation, when backpressure is propagating through the queue we can run into situations where shutdown can block for a long time. Since the shared groups semantics allow for us to release the received events, I think it makes sense to release the events that have not yet been accepted by the queue as quickly as we can notice it.

Maybe something like:


```suggestion
                if stop?
                  consumer.acknowledge(record, ack_release)
                else
                  handle_record(record, codec_instance, logstash_queue)
                  consumer.acknowledge(record, ack_accept)
                end
```

(assuming `ack_release` is a local var referencing `org.apache.kafka.clients.consumer.AcknowledgeType::RELEASE`)

lib/logstash/inputs/kafka.rb
```diff
@@ -294,15 +325,37 @@ def initialize(params = {})
...
     reassign_dns_lookup
     @pattern ||= java.util.regex.Pattern.compile(@topics_pattern) unless @topics_pattern.nil?
-    check_schema_registry_parameters
 
-    set_group_protocol!
+    if @consumer_mode == "share_group"
+      validate_share_group_config!
+      ack_type_class = org.apache.kafka.clients.consumer.AcknowledgeType
+      @share_group_ack_error_type = @share_group_acknowledgement_on_error == "reject" ?
+        ack_type_class::REJECT : ack_type_class::RELEASE
```
yaauie: 🤔 the line-break in the middle of the ternary makes it a bit easy to lose track of what is going on. This is definitely a nitpick, but I'd prefer to [leverage the if/else expression to assign the value](https://rubystyle.guide/#use-if-case-returns):


```suggestion
      @share_group_ack_error_type = 
        if @share_group_acknowledgement_on_error == "reject"
          org.apache.kafka.clients.consumer.AcknowledgeType::REJECT
        else
          org.apache.kafka.clients.consumer.AcknowledgeType::RELEASE
        end
```

lib/logstash/inputs/kafka.rb
```diff
@@ -294,15 +325,37 @@ def initialize(params = {})
...
+    else
+      check_schema_registry_parameters
```
yaauie, line 364: In the new `consumer_mode => share_group` we validate to make sure that the config doesn't invoke options that don't apply. I'd like to see the inverse, too. 

If a user specifies `consumer_mode => consumer_group` (or _effectively_ specifies it by leaving it undeclared and relying on the default value), I would consider it a configuration error to specify `share_group_acknowledgement_on_error`.

We can repeat the pattern from `SHARE_GROUP_INCOMPATIBLE_OPTIONS` and `original_params.keys`, just with a different list.

docs/input-kafka.asciidoc
```diff
@@ -53,13 +53,49 @@ The Logstash Kafka consumer handles group management and uses the default offset
...
+
+Requirements:
+
+* Kafka brokers 4.0 or later with `group.share.enable=true`
+* The `group_id` setting identifies the share group name
+
+Share group consumers must acknowledge each record. Successful records are always acknowledged
+as `ACCEPT`. The <<plugins-{type}s-{plugin}-share_group_acknowledgement_on_error>> option
+controls acknowledgement for records that fail processing (`release` for redelivery or `reject`
+to permanently discard).
```
yaauie: 🤔 within the context of a logstash pipeline, the input plugin only has visibility to the point at which the event is accepted by the pipeline's queue (and the queue is designed to block-until-accepted). I'm worried that this wording will lead users to believe that we can release/reject based on processing that happens downstream of the pipeline's queue.


```suggestion
Share group consumers must acknowledge each record.
After an event is accepted by the pipeline's queue, it is acknowledged as `ACCEPT` to ensure that no other consumer in the share group receives it.
When an event is either _not_ accepted by the pipeline's queue or an error occurs before it is offered to the queue, the <<plugins-{type}s-{plugin}-share_group_acknowledgement_on_error>> option controls how the event is acknowledged (`release` for redelivery or `reject` to permanently discard).
```

docs/input-kafka.asciidoc
```diff
@@ -285,14 +323,46 @@ https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+
...
+`group.share.enable=true` set in the broker configuration.
+
+[NOTE]
+====
+Setting any of the following options in the pipeline config raises a configuration error in
+`share_group` mode, since they do not apply to the Share Group protocol: `group_protocol`,
+`group_instance_id`, `partition_assignment_strategy`, `heartbeat_interval_ms`,
+`session_timeout_ms`, `isolation_level`, `enable_auto_commit`, `auto_commit_interval_ms`,
+`auto_offset_reset`, `check_crcs`, `exclude_internal_topics`, `schema_registry_url`.
+====
```
yaauie, line 371: I would prefer to add a note to each setting that only applies to a specific `consumer_mode`, so that a user looking at a given option can see whether it applies to them (instead of requiring them to cross-reference this)

> ~~~ asciidoc
> [NOTE]
> =====
> This option is supported only with `consumer_mode => consumer_group`.
> =====
> ~~~

docs/input-kafka.asciidoc
```diff
@@ -285,14 +323,46 @@ https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+
...
 
   * Value type is <<number,number>>
   * Default value is `1`
 
-Ideally you should have as many threads as the number of partitions for a perfect
-balance — more threads than partitions means that some threads will be idle
+In `consumer_group` mode, you should ideally have as many threads as the number of partitions
+for a perfect balance — more threads than partitions means that some threads will be idle.
+This does not apply in `share_group` mode, where any thread can fetch from any partition, so
+the number of threads is not bounded by the partition count
```
yaauie: Let's add some more context :)


```suggestion
The number of consumer threads this input should spawn for fetching events and pushing them into the pipeline's queue.

When configured with `consumer_mode => consumer_group`, each partition in the topic can only be assigned to a single consumer at a time, so if the _total_ number of consumer threads across all consumers in the group exceeds the number of partitions, some of those threads will be idle.

This constraint doesn't apply when configured with `consumer_mode => share_group` which is not partition-oriented.
```

docs/input-kafka.asciidoc
```diff
@@ -825,14 +895,32 @@ Security protocol to use, which can be either of PLAINTEXT,SSL,SASL_PLAINTEXT,SA
...
+* Only applicable when <<plugins-{type}s-{plugin}-consumer_mode>> is `share_group`
+
+Controls how a record is acknowledged when an error occurs while processing it.
+
+`release` (default): The record is released back to the share group for redelivery to any
+available consumer. Use this when processing errors are transient and the record should
+be retried.
+
+`reject`: The record is permanently discarded and will not be redelivered. Use this when
+the record is malformed and retrying would never succeed.
```
yaauie: 🤔 the guidance here assumes that the user will have insight into the _reason_ why processing failed (on a per-event basis), but since this is a general config I don't know if it makes sense to expose it at that level.

In the context of a logstash pipeline we have two situations where a record we have retrieved would require a non-ACCEPT ack:
 - an error occurs before we offer the event to the pipeline's queue (maybe malformed, crashing the codec?)
 - an error occurs while we're blocked by the pipeline's queue (very rare, likely only in interrupt/shutdown; the queue always accepts eventually but can block for a long time if the downstream pipeline workers aren't picking up work fast enough)

The trouble is we don't have a way to differentiate between them; they're both caught by the same `rescue` clause. We could _maybe_ alter the `handle_record` method to differentiate and route accordingly, but then we likely wouldn't need the config option.

```suggestion
Controls how a record is acknowledged when the event is either not accepted by the pipeline's queue or an error occurs before it is offered.

* `release` (default): The record is released back to the share group for redelivery to any available consumer. This strategy favors consistency over availability, which can allow a single malformed event to be retried indefinitely.
* `reject`: The record is permanently discarded and will not be redelivered. This strategy favors availability over consistency and risks data loss.
```

yaauie (commented):
Overall, excellent work and a welcome contribution. Thank you!

I've left a few comments in-line, addressing:

 - translating the kafka-ish docs to how the settings apply in the context of a logstash pipeline
 - validating that `share_group`-specific config options aren't used in conjunction with the `consumer_mode => consumer_group` (an inversion of the check already added that ensures that `consumer_group`-specific settings aren't used with the new `consumer_mode => share_group`)

docs/input-kafka.asciidoc
alexcams: Some of them are supported by share_group mode as well, but configured at broker/group. [KIP](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka#KIP932:QueuesforKafka-Groupconfiguration)
It may be worth clarifying this to be synced with Kafka docs.
| Consumer group (client-side) | Share group equivalent (broker/group-level) |
|---|---|
| `auto.offset.reset` | `share.auto.offset.reset` |
| `heartbeat.interval.ms` | `share.heartbeat.interval.ms` |
| `session.timeout.ms` | `share.session.timeout.ms` |
| `isolation.level` | `share.isolation.level` |

lib/logstash/inputs/kafka.rb
```diff
@@ -440,46 +508,63 @@ def maybe_commit_offset(consumer)
...
+
   def create_consumer(client_id, group_instance_id)
     begin
-      props = java.util.Properties.new
+      props = build_common_props(client_id)
       kafka = org.apache.kafka.clients.consumer.ConsumerConfig
 
       props.put(kafka::AUTO_COMMIT_INTERVAL_MS_CONFIG, auto_commit_interval_ms.to_s) unless auto_commit_interval_ms.nil?
       props.put(kafka::AUTO_OFFSET_RESET_CONFIG, auto_offset_reset) unless auto_offset_reset.nil?
       props.put(kafka::ALLOW_AUTO_CREATE_TOPICS_CONFIG, auto_create_topics) unless auto_create_topics.nil?
```
alexcams: [`auto_create_topics`](https://github.com/apache/kafka/blob/4.2/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ShareConsumerMetadata.java#L51) is supported by  `KafkaShareConsumer`. I tested it manually and works as expected (auto creates topic on `true` and don't on `false`).
I'd move it to `build_common_props`.

lib/logstash/inputs/kafka.rb
```diff
@@ -294,15 +325,37 @@ def initialize(params = {})
...
+    group_protocol
+    group_instance_id
+    partition_assignment_strategy
+    heartbeat_interval_ms
+    session_timeout_ms
+    isolation_level
+    enable_auto_commit
+    auto_commit_interval_ms
+    auto_offset_reset
+    check_crcs
```
alexcams: [`check_crcs`](https://github.com/apache/kafka/blob/4.2/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ShareFetchConfig.java) is supported as well. It is mentioned in [KIP](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka) at `Handling bad records` section.
It would need to be removed from this list and moved to `build_common_props` function.

orangecola:
@yaauie / @alexcams updated with suggestions, bumping for review.

PR #174, opened 2024-07-16
Update input-kafka.asciidoc

jguay:
Specify that only headers with UTF-8 encoded values are supported

PR #278, opened 2026-09-01
Add message_headers_field option to kafka output

### LINKED ISSUE #284: Kafka output: forward message headers from an event field (dynamic header names)
## Problem

The kafka output can only add headers whose **names are declared statically** in `message_headers`:

```ruby
message_headers => {
  "traceparent" => "%{[@metadata][kafka][headers][traceparent]}"
  "updateMode"  => "%{[@metadata][kafka][headers][updateMode]}"
}
```

When Logstash is used as a relay between a source topic and a destination topic (key, payload and headers preserved), the header set is not known in advance: producers add headers, different applications use different ones, and the list evolves over time. Any header not declared in the pipeline is read by the input (`decorate_events => extended`) and silently not forwarded by the output, and every new header requires a configuration change and a pipeline redeploy. Logstash should not need to understand the headers, only carry them.

When `message_headers` was introduced in #162, dynamic header names were deliberately left out pending a concrete use case. Mirroring the headers of consumed records is that use case. The same need shows up on the forum ([2021](https://discuss.elastic.co/t/writing-kafka-message-headers-using-logstashs-kafka-output-plugin/268625), [2023](https://discuss.elastic.co/t/creating-headers-with-key-value-in-kafka-offset-using-logstash-configuration/336089)) and in logstash-plugins/logstash-output-kafka#224.

## Proposal

A `message_headers_field` option referencing an event field that contains a hash. Each entry of that hash is added to the record as a Kafka header, after the static `message_headers`:

```ruby
output {
  kafka {
    topic_id              => "destination"
    message_key           => "%{[@metadata][kafka][key]}"
    message_headers_field => "[@metadata][kafka][headers]"
    codec                 => plain { format => "%{message}" }
  }
}
```

- Values are used verbatim (`to_s`), no `%{}` interpolation, since they are data rather than templates.
- Absent field → event sent without extra headers; non-hash field or non-scalar values → skipped with a warning logged once; the event is never dropped.
- Invalid or empty field reference → `LogStash::ConfigurationError` at `register` time.
- `message_headers` is unchanged; both options can be combined.

Implementation with unit specs, an integration spec against a real broker, and docs: #278.

mbelaloui:
Closes #284

## Summary

Adds a `message_headers_field` option to the kafka output. It references an event field containing a hash, and each entry of that hash is added to the produced record as a Kafka header — so a set of headers that is not known in advance can be forwarded, typically the headers of a consumed record captured by the kafka input's `decorate_events => extended` mode.

```ruby
output {
  kafka {
    topic_id              => "destination"
    message_key           => "%{[@metadata][kafka][key]}"
    message_headers_field => "[@metadata][kafka][headers]"
    codec                 => plain { format => "%{message}" }
  }
}
```

## Motivation

In our architecture the Kafka payload is the business object and its schema stays limited to functional data. Technical metadata — correlation id, distributed tracing context (`traceparent`), processing directives such as an update mode, and monitoring information — travels in Kafka headers. This keeps the business schema clean, avoids coupling the domain model to tracing/monitoring concerns, and lets technical metadata evolve without a schema or version change on every producer and consumer.

Logstash is used as a transparent relay between a source topic and a destination topic (key, payload and headers preserved). The kafka input already exposes the record headers under `[@metadata][kafka][headers]`, but the output can only add headers whose **names are declared statically** in `message_headers`:

```ruby
message_headers => {
  "breadcrumbId" => "%{[@metadata][kafka][headers][breadcrumbId]}"
  "updateMode"   => "%{[@metadata][kafka][headers][updateMode]}"
  "traceparent"  => "%{[@metadata][kafka][headers][traceparent]}"
}
```

This works while the header list is fixed, but breaks down as soon as producers add headers, different applications use different headers, or the list evolves over time: any header not declared in the pipeline is read by the input and silently not forwarded, and every new header requires a configuration change and a pipeline redeploy. Logstash should not need to understand the headers — it only needs to carry them.

When `message_headers` was introduced in #162, header names were deliberately kept static pending an explicit use case for dynamic names. Mirroring the headers of consumed records is that use case; the need also shows up on the forum ([2021](https://discuss.elastic.co/t/writing-kafka-message-headers-using-logstashs-kafka-output-plugin/268625), [2023](https://discuss.elastic.co/t/creating-headers-with-key-value-in-kafka-offset-using-logstash-configuration/336089)) and in logstash-plugins/logstash-output-kafka#224.

## Behavior

- Headers from the referenced hash are appended **after** the ones from `message_headers`. Kafka allows several headers with the same name, so a name present in both results in two headers on the record (same semantics as `Headers.add` in the Java client).
- Header names are the hash keys; header values are `value.to_s` converted to bytes with the same `to_java_bytes` call as the existing `message_headers` path. Values are used verbatim — no `%{}` interpolation is applied, since dynamic values are data rather than templates and may legitimately contain `%{...}` sequences. Entries with an empty key or a `nil` value are skipped silently; non-scalar values (nested hashes/arrays) are skipped with a warning, logged once per plugin instance.
- If the referenced field is absent the event is sent without additional headers. If it does not contain a hash, a warning is logged once and the field is ignored. The event is never dropped because of this option.
- The field reference is validated at `register` time: an invalid or empty reference raises a `LogStash::ConfigurationError` at startup instead of failing on every event. Validation is done manually rather than via the core `:validate => :field_reference`, because the core validator silently treats an empty or whitespace value as unset, whereas an explicit error surfaces the misconfiguration.
- `message_headers` is unchanged and both options can be used together.

## Changes

- `lib/logstash/outputs/kafka.rb`: new option, register-time validation, `add_dynamic_message_headers` called in `write_to_kafka` right after the static headers.
- `spec/unit/outputs/kafka_spec.rb`: 9 unit specs (hash forwarded with scalar values coerced via `to_s`, key/value untouched, absent field, non-hash field, empty key / nil value, non-scalar values, static+dynamic duplicate names, invalid and empty reference).
- `spec/integration/outputs/kafka_spec.rb`: integration case mirroring the existing `message_headers` one — headers read from the referenced field are asserted on records consumed from a real broker.
- `docs/output-kafka.asciidoc`: option added to the table and documented.

CHANGELOG and `version` are intentionally left untouched for the `/pr_prepare` workflow. Suggested entry (minor): `Add \`message_headers_field\` option to forward Kafka headers from an event field in the output plugin`.

## Testing

- Unit suite green on this branch: 87 examples, 0 failures (`spec/unit`), including the 9 new examples in `kafka_spec.rb` (23 before, 32 after).
- End-to-end with a gem built from this branch installed on Logstash 8.19.11, mirroring a source topic to a destination topic on a Kafka 4.3 broker: key, headers (`breadcrumbId`, `updateMode`, `traceparent`) and payload preserved; a record without headers is mirrored without headers.

## Notes

Header fidelity is bounded by what the input captures with `decorate_events => extended`: values that are not valid UTF-8 are skipped (the limitation documented in #174), duplicate names collapse to the last value, and header order is not preserved (headers are exposed as a hash keyed by name). This option forwards what is present in the event and does not change the input.


mbelaloui:
/pr_prepare release:minor
- Add `message_headers_field` option to forward Kafka headers from an event field in the output plugin

mbelaloui:
Hi @mashhurs @yaauie — since you both reviewed #162, which introduced `message_headers`, would you have a moment to look at this follow-up? It covers the dynamic-header-names case that was left open there (use case and context in the linked issue).

The branch is rebased on `main` and includes #282; the CI workflows still need a maintainer's approval to run. Thanks!

mashhurs:
@mbelaloui I will discuss with team this but if you are blocked by this issue, one workaround I do see is using ruby-filter before kafka-output:
```ruby
filter {
  ruby {
    code => '
      headers = event.get("[@metadata][kafka][headers]")
      if headers.is_a?(Hash)
        headers.each { |k, v| event.set("[message][headers_out][#{k}]", v) }
      end
    '
  }
}
```

PR #282, opened 2026-09-16
[Test] create bigger payloads to pass Kafka batch size

andsel:
During Logstash output Kafka send bigger events so that the Kafka's producer batch size is passed to force the partitioner to kicks in.

- Closes #281

PR #283, opened 2026-09-17
[Backport 11.x #282][Test] create bigger payloads to pass Kafka batch size

andsel:
During Logstash output Kafka send bigger events so that the Kafka's producer batch size is passed to force the partitioner to kicks in.


andsel:
Close as not applicable to version `11.x` because uses the `UniformStickyPartitioner` which stick to a partition till the batch size or linger_ms are passed. Once it switch to a new partition, it doesn't apply a round robin policy but select one randomly so there isn't any certainty that about the equal distribution on the 3 paritions.

